PR #3, opened 2021-05-21
Global config

nmn:
Things in the Pull Request:

## moved to using just NPM, no more yarn
In order to simplify setup, I have removed the dependency on Yarn. You can now just use NPM for everything.

## Moved dependencies and build steps to top-level
All the packages, set up and dependencies have been moved to the top-level of the repo. This means that there is no need to install the same packages for every plugin.

## Every package gets linting and prettier
All javascript files (except those that are explicitly ignored in `.eslintrc`) will now get linting with ESLint and auto formatting with Prettier. Hopefully, no one has strong opinions on code style.

I have gone and edited all the JS files to fix every issue that arose.

## Optional typechecking
If you add the `// @flow` comment at the top of your JS file, you will also get Flow typechecking, autocomplete etc.

A couple of files were already using this and I went and fixed all the type errors.

## Bundling
`nmn.sweep` uses `rollup.js` to bundle a folder called `src` into a single file and automatically remove flow type annotations.
In order to simplify usage for others, I will create a script that looks in all the plugin folders and automatically do the same bundling if  the file `index.js` in a sub-folder called `src` is found.

## Outcome
1. Code-style consistency with linting and auto-code-formatting.
2. Code correctness with optional type checking that is super easy to add.
3. Dead-simple setup to get type annotations and bundling.


jgclark:
Thanks, @nmn, for all this. 
The eslint settings mean almost every line of @EduardMe's code and mine has changed, if only to add closing `;` and change `"` to `'`. I picked up my style from Eduard's original examples script, but I'm happy with this style change if that's in the direction of more idiomatic JavaScript.  (And when writing new code, I've defaulting to declaring new variabales using `let` unless I know they're intended to be `const`. But amazing the number that don't then change, that eslint has spotted.)

I'm now going to attempt to apply the PR locally to test it hasn't broken anything.

jgclark:
I've noticed one problem: something (presumably eslint?) has changed a `"double-quoted string than included \n characters"` into a string enclosed by three backticks. This it didn't like at all. It seems they to be single backticks.

nmn:
> I've noticed one problem: something (presumably eslint?) has changed a `"double-quoted string than included \n characters"` into a multi-line string enclosed by three backticks, which made it clearer to read. However it failed to run. It seems they to be single backticks.

Yup. That's my mistake. I'll fix it. Good catch!

jgclark:
Thanks.
Well, I'd happily switch back to not using semicolons.  Missing some off accidentally causes a script to fail silently in my experience.  They feel unnecessary cruft, particularly when coming from Ruby, which feels very clean and lean in comparison.

I'm looking forward to these checking systems catching more errors, like the ones you fixed.

jgclark:
BTW How do I mark this PR as reviewed, without actually doing the Merge?

nmn:
> BTW How do I mark this PR as reviewed, without actually doing the Merge?

I don't know...

jgclark (approved):
Now found the right button to approve without merging, I hope.

PR #9, opened 2021-05-26
Starting work on new Templates Plugin

nmn:
I have almost the whole thing working except parsing data format options. I'm considering a few options for that.

But here's how it works so far:

ezgif-3-9793ea9c29c0

This is what I'm thinking:
1. A root folder called Templates that holds user defined templates.
2. Those templates can have {{smart-value}} tags within.
3. The plugin would let you easily create a new note with a given template.
4. Some of those values could be smart, e.g. {{date dd-mm-yyyy}}  and would automatically be replaced with the correct date with the given format.
5. All other tags would be filled by asking the user for input when the plugin is running.
6. There will also be a note within the Templates folder called Config that will have JSON configuration (wrapped within codefences). Any keys defined within that config will automatically be used.


brokosz:
Nice! Saved me a bunch of coding to build this myself this weekend. :D  It would be great to have a function to take an existing note and copy it into a new template. I was thinking the root folder should be outside the Notes folder, at the same level with Calendar, Themes, and Plugins. But having it under Notes would make them easier to edit later.

jgclark:
@nmn, some comments, before looking at the code:

- I'm happy to shift out some of the Template-ing functionality to a separate Plugin specifically for that. I figured we'd do that sooner or later.
- I think the Obsidian Templater referred to by @dwertheimer should be a good guide to the syntax:
  - For example it distinguishes between variables and functions. We'll want that, I'm sure.
  - It also has a three part naming structure. That may be overkill, but perhaps a 2-part one would have its uses.
  - For example, the way to do yesterday's date in that scheme is ` <% tp.date.now("YYYYMMDD", -1) %>`
- As Templates are for any user, not just an advanced thing for coders, using JSON doesn't feel ideal.  We've had discussions elsewhere about JSON vs TOML: I suggest this would be the place to try TOML instead.

I also agree with @brokosz that the Templates should ideally be stored off the root level, but that obviously needs a structural change from @EduardMe.

jgclark (dismissed):
OK, now I've looked at the code, I see that you're suggesting different command names but for very similar commands:
- /addTemplate looks to be the same as my existing /it
- /newNoteWithTemplate might or might not be as functional as my existing /nn (extended by Eduard).

We decided about a week ago to go for short command names, but with useful descriptions.

nmn:
> OK, now I've looked at the code, I see that you're suggesting different command names but for very similar commands:
> 
> * /addTemplate looks to be the same as my existing /it
> * /newNoteWithTemplate might or might not be as functional as my existing /nn (extended by Eduard).
> 
> We decided about a week ago to go for short command names, but with useful descriptions.

I purposely kept a temporary name because I didn’t want to create conflicts. I was not aware that we decided on short names for plugins. I find that very confusing personally but I can live with it for now.

nmn:
> @nmn, some comments, before looking at the code:
> 
> * I'm happy to shift out some of the Template-ing functionality to a separate Plugin specifically for that. I figured we'd do that sooner or later.
> * I think the Obsidian Templater referred to by @dwertheimer should be a good guide to the syntax:
>   
>   * For example it distinguishes between variables and functions. We'll want that, I'm sure.
>   * It also has a three part naming structure. That may be overkill, but perhaps a 2-part one would have its uses.
>   * For example, the way to do yesterday's date in that scheme is ` <% tp.date.now("YYYYMMDD", -1) %>`
> * As Templates are for any user, not just an advanced thing for coders, using JSON doesn't feel ideal.  We've had discussions elsewhere about JSON vs TOML: I suggest this would be the place to try TOML instead.
> 
> I also agree with @brokosz that the Templates should ideally be stored off the root level, but that obviously needs a structural change from @EduardMe.

Just noticed this comment!

- I’ll update this PR to rename the command to /it and remove your existing plugin that does the same thing.
- I generally like the idea of Obsidian Templater but I had a few issues:
  - I could support date with a custom format like YYYYMMDD, but it also seemed useful to have support for timezones. I’ll try to find a way to add both.
  - I didn’t want to support custom JavaScript functions as I want to avoid running untrusted code for now.
  - I do want ideas for functions that can be built in.
  - I can also add parsing to accept arguments to those built-in functions that I planned on adding from the start.
- About JSON vs TOML I had that discussion and I think we settled on JSON5. It has comments and other features we wanted but is still valid JS and more readable for complex configuration. Read more at https://json5.org/

jgclark:
Thanks for your responses, @nmn. All's good, I think.

I wasn't trying to suggest limiting ourselves to Obsidian Templater, but where we want to provide the same functionality, we might as well use their schemes.  But to add or extend where needed, for example to allow timezones etc.

An example of a function to build in is my weather lookup. Once we have this sort of mechanism in place I was going to make it more flexible, in the same we can specify date output formats.

nmn:
A weather function sounds like a great idea. I’ll make it accept a city name, or latitude and longitude along with a date and hook it all up.

jgclark:
> A weather function sounds like a great idea. I’ll make it accept a city name, or latitude and longitude along with a date and hook it all up.

Great. The OpenWeather API requires an API Key (which could be specified in the Template that calls the function) and an optional 'units' ('metric', 'imperial' from memory).

dwertheimer:
https://openweathermap.org/api/one-call-api

EduardMe:
@nmn looks great!

We could update the /it command with this.

We have created some code to automatically create a templates folder upon
first usage. It also creates two sample template notes. And you can choose
the template folder (the sub folder where you want to keep the templates
folder. The plug-in later searches for it in all sub folders by name).

I think this would help users understand how the plug-in works without
having to read a readme and know where the template folder is located.

Check it out!

nmn:
Here's a quick status update. The plugin has the following features working:
1. The first time you try to insert a template, it creates `Templates` folder and adds a few templates to get you started.
2. It also adds a file called `_configuration` (we could change this?) where you can configure default values for the plugin.
    * You can use JSON, JSON5, YAML or TOML for your configuration
    * The dummy file gives you examples and instructions for all
    * In the future, it'll read your weather API key from this config file.
3. When inserting a template, it looks for anything wrapped within `{{` and `}}` (handlebars) and treats it as a tag.
    * If it detects a known function call within a tag, like `date()`, it uses the built-in date function.
    * You can even pass a single object parameter to the function to customize the way it behaves. I use the JSON5 parser to parse the argument and use the values.
    * For any unknown tag, it asks the user for input. Making it work like editor snippets.
    
Next Steps:
1. Make the `new note` command work as well
2. Replace the existing "Templates" plugin with this by renaming the command name.
3. In the near future, add the built-in weather function as well. Weather dates are little complex to figure out.
    
Here's a few screenshots to show how things work:

This is the original Template File:
Screen Shot 2021-05-29 at 12 22 21 AM

This is my global Configuration:
Screen Shot 2021-05-29 at 12 22 32 AM

This is what was inserted in my note:
Screen Shot 2021-05-29 at 12 22 38 AM

Only `mediaName` was an unknown value and was asked for during Template inserting. Everything else was resolved automatically.

dwertheimer:
@nmn This is very cool. Coming along!
In case it's helpful, here's the way I modified Jason's weather code for my own use. Simple, but might be useful to you in the plugin as a default response:

`  if (pref_openWeatherAPIKey != "") {
    var getWeatherURL = "https://api.openweathermap.org/data/2.5/onecall?lat=" + pref_latPosition + "&lon=" + pref_longPosition + "&exclude=current,hourly,minutely&units=" + pref_openWeatherUnits + "&appid=" + pref_openWeatherAPIKey
    console.log(getWeatherURL);
    fetch(getWeatherURL)
      .then(jsonIn => {
        var weatherResponse = JSON.parse(jsonIn);
        var weatherMyLocation = weatherResponse.timezone;
        var units = pref_openWeatherUnits === "imperial" ? "°F" : "°C"; // Who checks the weather in Kelvin? :)
        var weatherTodayAll = weatherResponse.daily["0"]
        var maxTemp = weatherTodayAll.temp.max.toFixed(0)
        var minTemp = weatherTodayAll.temp.min.toFixed(0)
        var weatherDesc = weatherTodayAll.weather["0"].description
        var weatherLine = `Weather: ${minTemp}${units}-${maxTemp}${units} / ${weatherDesc} in ${weatherMyLocation}`;`

dwertheimer:
@nmn Minor typo in your sample config file:
lattitude => latitude

Also, people may not know how to easily get their Lat/Long. Maybe you add a comment:
```
# To get weather, you will need the Latitude and Longitude of your city or your specific location (e.g. your home)
# If you don't know your Latitude/Longitude, go to https://www.latlong.net/ and type in your city or full address
# Enter the results here:
```

nmn:
@dwertheimer nice catch on those typos. I’ve reading the weather API and I plan to add support for just taking a city name as well.
The Latitude and longitude will be the backup.

EduardMe:
@nmn <https://github.com/nmn> Looks awesome! Is the configuration file a
note? If yes, ignore below.

 I know ideally this should be configurable later in the preferences. It
could just take the TOML/JSON and reproduce controls for each line later,
but right now the easiest way to modify a configuration file would be to
change the text in a code block, which is in a note I guess. It could sit
in the same Templates folder.

nmn:
> Is the configuration file a note? 

Yup it's just a note. Let me share how it all works:



(I noticed that the Date tag failed to work, so I'll fix that in a commit now)

nmn:
Date issues fixed.

brokosz:
Is there a way to get ISO 8601 formatted dates `YYYY-MM-DD` or do I need to edit the plugin to do that?

dwertheimer:
Brad's Q: Can I get an ISO8601 date, like 2021-06-21, using Templates?
A: Well, Brad, the Swedes are your friend, because the country was nice
enough to adopt 8601 for dates.

So, thanks to the Swedes, here's how you can get an 8601 with no code
editing:

1) edit your _configuration  file in Templates and comment out or delete
the `timeStyle` line altogether (if there's a timeStyle, it will always
return a time, even if you don't want one)
2) create your date templates like this:
`Date using Swedish Locale  {{date({locale: 'sv-SE', dateStyle: 'short'})}}
// yields 2021-06-21`
…and that answers your original question, but if you ever want the time
also, add back in a timeStyle in the tag:
`Date using Swedish Locale with time  {{date({locale: 'sv-SE', dateStyle:
'short', timeStyle:'short'})}} // yields 2021-06-21 22:21`
…so far so good. But if you want that to read 10:21 PM, we're gonna need a
code change. I could almost make it work by commenting out the dateStyle in
the _configuration file (so it wouldn't by default always give me a date)
and then adding a second tag for the time:
`{{date({locale: 'en-US', timeStyle: 'short', hour12:true})}} // should
yield 10:21 PM`
…doesn't quite work because the Template code in interpolation.js forces a
dateStyle override of 'en-US' if you don't put one in, and once you do
that, you get a date in US format with your time: `6/21/2021 10:21 PM`,
even though you only wanted the time part.

If @nmn is open to revisiting the following override in
interpolation.js::processDate, we could get there, by having two tags, one
for 8601-date and one for time.
`const secondParam = {
dateStyle: 'short',
...otherParams,
}`

However, @jgclark and @nmn, I would recommend we add an `8601date` Tag that
will provide dates in this format without all the hoop-jumping. We can also
add a "time" tag, but that will require the secondParam change above if we
want to use the IntlDate and processDate code.

brokosz:
Wow, that was thorough! Thanks @dwertheimer!!

PR #12, opened 2021-06-11
/nns - New Note From Selection

dwertheimer:
@jgclark Pls have a look. Thanks @EduardMe for the suggestion -- I decided to make sure I am creating a unique file title before I create the link.


jgclark:
You're almost there:
- Please could you bump version to v0.9.0 in header comments and plugin.json?
- plugin.json also needs minimum version 3.0.23 for the new functions you're using
- .projectNoteByTitle used to have 2 parameters, now has 3, and your use has 1. I think you want to add ( ..., true, false) but you'll need to check that.

Thanks.

jgclark:
PS I'm just about to commit a change to that plugin's README for you, which is also needed.

dwertheimer:
Great. Thanks for the feedback and for helping me find some of these nuances. I will keep an eye out for the updated readme.

dwertheimer:
Thanks again @jgclark 
Hopefully this will do it. 

BTW, you uncovered an inconsistency in the documentation:
https://help.noteplan.co/article/70-javascript-plugin-api#datastore

The documentation has an example which uses one param:
`// var note = DataStore.projectNoteByTitle("TEST")[0]`

jgclark (approved):

jgclark:
Reading the code looked fine, so merged (as above).
Now I've tested the code, it's not working: on first four tests it's always losing the selected text, apart from the first line, which does get made into the title of a new note.
So most of is OK, but the only logging I get is `newNoteFromSelection (running):`.

jgclark:
> BTW, you uncovered an inconsistency in the documentation:
> https://help.noteplan.co/article/70-javascript-plugin-api#datastore
> 
> The documentation has an example which uses one param:
> `// var note = DataStore.projectNoteByTitle("TEST")[0]`

@EduardMe: it looks to me that either your example needs a fix, or the documentation needs to be changed to show the second two parameters are optionals.

dwertheimer:
I will do some more testing. Thanks.
Are you testing it on the newest beta build?

jgclark:
Yes, I only report when running the latest builds of things.
Here's a movie of what happens, which is the same as I reported before. I've checked that this all from the main window, as I realised sub-windows might complicate things.

https://user-images.githubusercontent.com/1671754/121822423-ef7c6600-cc96-11eb-8920-56f9b0d8305e.mov

Log only shows `newNoteFromSelection (running):`.

dwertheimer:
That's fascinating. Have you pulled my code from my latest branch with the
open PR (0.9.1)?
I will do some more testing. I will also add more console.logging.
Thanks Jonathan!

jgclark:
Just for completeness, I now see your code works when creating note in the root folder. But not in sub-folders.  It looks to be that API ambiguity about when filename = filename, or when it = folder+filename.

EduardMe:
This should be hopefully auto-magically solved in the next beta update.
`newNote` was not caching the note when it was created in a folder.

PR #13, opened 2021-06-12
Correct missing single quote

dwertheimer:
This missing quote in the default _configuration setup was keeping _configuration from loading by the plugin (throws an error)


dwertheimer:
@jgclark Fixing a bug in the default template creation which caused the config file to not load. May send you some other suggestions, but wanted to get this bug fixed right away.

jgclark (approved):
Approving quickly, so we can make progress.

jgclark:
BTW, ... now I've looked more closely I see this is for the auto-generated script. This will get overwritten next time someone use the build system.
I will as well fix this in the underlying script from @nmn (configuration.js), and I'm in the process of raising a wider PR that will include this.  Please do get involved in that discussion ...

PR #14, opened 2021-06-12
Updates to Templates #11

jgclark:
Structure and readability changes only, I hope.


jgclark:
@dwertheimer may want to pitch in with ideas too?

nmn.Templates/src/interpolation.js
```diff
@@ -50,18 +49,44 @@ async function processTag(
   return await getInput(`Value for ${tag}`)
 }
 
+// Apply any matching tag functions
+export async function processTags(
```
nmn, line 53: This function is defined twice

nmn (changes requested):

nmn.Templates/src/interpolation.js
dwertheimer: Exists in `interpolation.js` and `tagFunctions.js`

jgclark:
Fixing multiple definition: going back to a single file 'interpolation.js' file, though I don't think the name is very clear. I was trying to make a smaller file where it was clearer that people should add new functions, but circular defintions stopped that.

(PS I'm finding my way around the GitHub extensions in VSC quite confusing for PRs, so apologies if I goof in the process.)

nmn.Templates/src/interpolation.js
jgclark: Thanks. Hopefully the next commit sorts this.

jgclark:
Not having this resolved was causing me increasing pain as it was blocking getting other code released. So, as I've not had any comments in 5 days, I've gone ahead and made the commits. 

The main changes are:
- moving my weather lookup code from DailyJournal to Templates plugin
- making it more obvious where to add new functions to Templates
- renaming 'addTemplate' function to 'applyTemplate'.
- adding 'applyNamedTemplate' function

PR #15, opened 2021-06-13
Writing to file off-screen so it displays upon open

dwertheimer:
Working around a bug/feature where the appendParagraph was not showing up in the Editor. It's better this way anyway, because I can ask the user whether to open it or not. :)

@jgclark pls let me know if this works better for you

### DIFF
--- a/jgclark.NoteHelpers/README.md
+++ b/jgclark.NoteHelpers/README.md
@@ -20,7 +20,7 @@ Templates should normally end with a linefeed character (`\n`).
 
 ## History
 
-### v0.9.1, 12.6.2021
+### v0.9.2, 13.6.2021
 - change: write text offscreen and ask user if they want to open newly created file
 
 ### v0.9.0, 12.6.2021
--- a/jgclark.NoteHelpers/README.md
+++ b/jgclark.NoteHelpers/README.md
@@ -20,6 +20,9 @@ Templates should normally end with a linefeed character (`\n`).
 
 ## History
 
+### v0.9.1, 12.6.2021
+- change: write text offscreen and ask user if they want to open newly created file
+
 ### v0.9.0, 12.6.2021
 - [add]: **/nns** command to add a new note from selection (and leave link to it in its place) (@dwertheimer)
 
--- a/jgclark.NoteHelpers/noteHelpers.js
+++ b/jgclark.NoteHelpers/noteHelpers.js
@@ -74,12 +74,13 @@ async function showMessageYesNo(message, choicesArray = ['Yes', 'No']) {
 // Find a unique note title/filename so backlinks can work properly (@dwertheimer)
 function getUniqueNoteTitle(title) {
   let i = 0,
-    res = []
+    res = [],
+    newTitle = title
   while (++i === 1 || res.length > 0) {
-    newtitle = i === 1 ? title : `${title} ${i}`
-    res = DataStore.projectNoteByTitle(newtitle, true, false)
+    newTitle = i === 1 ? title : `${title} ${i}`
+    res = DataStore.projectNoteByTitle(newTitle, true, false)
   }
-  return newtitle
+  return newTitle
 }
 
 //------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


jgclark (commented):
(I've now found how to check PRs without approving first.)

I've tried this, and I can't see an obvious problem, but it appears to be failing at line 282 `newNote.appendParagraph(movedText, 'empty')` for no obvious reason. I added some temporary logging and the `movedText` is OK, but it doesn't append anything, and code execution stops then.

Otherwise the only thing I can spot is that the `.shift()` is being done on a const array, though it doesn't seem to matter, for reasons which are beyond me.

dwertheimer:
I think shift() on a const array is ok because arrays in JS are all
mutable. To me that’s one of the confusing things about const with Objects
and Arrays. You can’t overwrite them entirely but you can edit the contents
(eg object properties) or array items at will.

In fact, the first example in the JS manual does exactly this:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift

jgclark:
> I think shift() on a const array is ok because arrays in JS are all mutable. To me that’s one of the confusing things about const with Objects and Arrays.

Thanks -- a bit confusing, but good to know!

jgclark:
I'm really willing this on to work. However, first test of this (to a sub-folder) fails:
```
newNoteFromSelection (running v0.9.2) 6 selected:
	1st Para Type = text = "one"
	currentFolder=TEST
	origFile:TEST repeats
	newNote returned Filename:TEST/one.md
	About to open filename: "TEST/TEST/one.md" (no leading slash) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "/TEST/TEST/one.md" (with leading slash) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "TEST/TEST/one.md" (Jonathan way) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "TEST/one.md" (no folder path, filename only) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "TEST/TEST/one.md" (using noteByFilename Instead) using noteByFilename
		Didn't work! noteByFilename returned undefined
	About to open filename: "/TEST/TEST/one.md" (using noteByFilename Instead, leading slash) using noteByFilename
		Didn't work! noteByFilename returned undefined
Could not open file: "/TEST/TEST/one.md"

newNoteFromSelection (finished)
```
Second try, to root, also fails:
```
newNoteFromSelection (running v0.9.2) 6 selected:
	1st Para Type = text = "one"
	currentFolder=/
	origFile:TEST repeats
	newNote returned Filename:one.md
	About to open filename: "one.md" (no leading slash) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "one.md" (with leading slash) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "//one.md" (Jonathan way) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "one.md" (no folder path, filename only) using projectNoteByFilename
		Didn't work! projectNoteByFilename returned undefined
	About to open filename: "one.md" (using noteByFilename Instead) using noteByFilename
		Didn't work! noteByFilename returned undefined
	About to open filename: "one.md" (using noteByFilename Instead, leading slash) using noteByFilename
		Didn't work! noteByFilename returned undefined
Could not open file: "one.md"

newNoteFromSelection (finished)
```

I'll now try and work on it myself.

dwertheimer:
Sorry Jonathan. @Eduard said he was going to look at the API today. My
code may be buggy, but the docs should be more clear on how “filename”s
work. Might be a good idea to use the parameter “path” to delineate from
“filename” in the appropriate APIs.

dwertheimer:
Hey this may be a clue in your output:

newNote returned Filename:TEST/one.md About to open filename:
"TEST/TEST/one.md"

I have never seen addNote return a directory-slash-filename (path). Is this
new? Why did this happen here and not for us before?

dwertheimer:
Have just mentioned it on Discord. There was a bug where the notes which
were created inside a folder with `newNote` weren't cached immediately. So
the returned filename couldn't be used to query the note unless you had an
await CommandBar user input in between. Will be fixed in the next beta.

@David, you are right about the confusion with the filename. Actually, it
should have been a relative path all along. So yes, since the last Beta
update `newNote` returns the relative path. I have updated the API
documentation to point out that the relative folder is included in returns
and must be in function variables.

dwertheimer:
Closing PR because moving code to Filer

PR #18, opened 2021-06-15
Moved /nns to Filer

dwertheimer:
Now that /nns works, for modularity and ease of maintenance, decided to break out /nns and put it in Filer.


dwertheimer:
@jgclark Let me know what you think. I tested both and both seem to work properly.

dwertheimer:
Sorry, tried to rename the branch, but that deleted my PR

jgclark:
I've made mods to quite a few files, principally to add @flow typing, and to correct a stray comma in NoteHelpers/plugin.json that would stop it working. 
I've pushed two commits to this PR branch, but I can't see them appearing -- argh!

PR #20, opened 2021-06-19
Created a few date plugins for things I do frequently (with some help from code by @nmn

dwertheimer:
/now - insert local date/time at cursor
/date - insert local date at cursor
/time - insert ISO format date/time at cursor
/ldn - insert link to the daily note at the cursor

Is a pull request like this the proper way to get something new into the plugins repo?


jgclark:
Yes, raising a PR like this is the right way to get new plugins added to the repo.
I've looked at the code, and I've got two suggestions:

1. Is `/time` the best name for one of them, given you're including date as well? (I hesitate to say this as naming is never easy ... particularly finding short names.)
2. People in different countries are likely to want to different formats for some of these. I see that in the Templates code that handles date tags, @nmn included a package from `Intl`. Perhaps if the locale string is set in the `_configuration` file, you could pick that up and use it?

Let me know if you want to make changes. If not then I can merge to main and create a release for you, so that it becomes visible in the app.

dwertheimer:
@jgclark Made all the dates and times conform to the user's specified formats in _configuration. Reads it if there is one.
I'm not creating a _configuration file if there isn't one -- you just get my defaults if you don't specify otherwise. Pls let me know what you think.

dwertheimer.DateAutomations/src/dateFunctions.js
```diff
@@ -0,0 +1,123 @@
...
+        text: `${text}`,
+      })
+    }),
+  )
+  // console.log(JSON.stringify(options))
+  return options
+}
+
+// /iso
+export async function insertISODate() {
```
nmn, line 71:
```
export async function insertISODate(): void {
```

nmn (approved):

jgclark:
Good work.
See my comments elsewhere about the extensions I'm making to add new default config data into _configuration. This is with others in mind, so you might want to pick this up when it's ready.

jgclark:
I hope that merge is OK. I'm working from GitHub's iPad client for the first time.

dwertheimer:
Yes, looking forward to Using that config code.

PR #22, opened 2021-06-26
Sweep Streamline (one file)/Extend (/sw7)

dwertheimer:
@nmn I made a lot of edits and clarifications. Todos in Project Notes weren't sweeping properly, so I dug in and decided to streamline the code, combining the Calendar Note and Project Note processing into one file, since it's 90% the same.  @EduardMe and @nmn Please try it out and see if it still works properly for you. I tested it out extensively, and everything works for me. 

- Fixed project note sweeping so tasks copy from project notes
- Combined Calendar and Notes processing into one file for consistency 
- Added /sw7 command to silently sweep notes from the last 7 days (sweep7() in sweepAll.js)
- Added comments for understanding the indents logic and console.logging
- Edited messaging for clarity for user
- Started fleshing out README

PR #23, opened 2021-06-27
Task sorter v0.0.2

dwertheimer:
Sort tasks in a note by priority (e.g. !, !!, !!! -- the more exclamations, the higher the priority -- also works with todo.txt (A), (B) etc.)
My personal workflow: Sweep tasks with /sw7 and then sort tasks with /ts (see my /sw7 pull request also)
Ignores indents for now, but once I understand more about @EduardMe's heading-indent thinking, I can get it back in there.

Need to work on the @context and #hashtag sorting next.

@jgclark @EduardMe Look forward to comments


EduardMe:
I think @dwertheimer, "indent" values are only available on tasks and bullets because these are the only items where you can indent the full paragraph. Plain text, headings, etc are not 'indentable' like tasks. You only get the first line indented. There is a discussion on Discord about indenting all fully. Then it would make sense to set an indent level everywhere.

Let me know if this helps or if I mixed up something.

dwertheimer:
Thx @EduardMe. When you get a moment, pls try out my PRs for the task sweeper and the sorter. 

Task sweeper sweeps the last #H1 over with an indented task. The sorter, by design, just sorts tasks and ignores text and titles.  So at some point, we should come up with a strategy for how to do the sorting when there are indents and titles. The base case is simple but let's say you sweep multiple indented tasks of differing priorities to today's note. They are all now under the #H1. But when you sort them, what should happen?

EduardMe:
Thanks, will check it out. Generally, indented tasks should stay below
their parent task, but that's not easy to implement (searching for the
parent) and then to sort indented tasks only between the other indented
tasks under the same parent. Because there might be tasks which are further
indented into level 2, 3, etc. I wanted to code this somehow, but couldn't
figure out a simple way to get the relationships right.

jgclark:
I've only checked which files have been updated (given the issues reported in #22). 
So not reviewed, or tested, but committing as requested.

PR #44, opened 2021-08-07
feat: add reminders plugin

weyert:
Introduces the Reminders plugn allows to insert a list of today's reminders into a daily note, or use within a Template file


nmn.Templates/src/templateController.js
```diff
@@ -26,6 +27,8 @@ addTag('matchingEvents', listMatchingDaysEvents)
 addTag('listMatchingEvents', listMatchingDaysEvents)
 addTag('quote', getDailyQuote, true)
 addTag('sweepTasks', sweepTemplate)
+addTag('reminders', listReminders)
```
nmn, line 30: I would prefer a single tag for this. At least until we can add some sort of forEach syntax.

weyert.ReminderHelpers/src/reminders.js
```diff
@@ -0,0 +1,368 @@
...
+
+  // If the date is not including - we assume its YYYYMMDD
+  console.log(`getDateFromString() Date string is missing -`)
+  const year = parseInt(dateStr.slice(0, 4))
+  const month = parseInt(dateStr.slice(4, 6))
+  const day = parseInt(dateStr.slice(6, 8))
+  return [year, month, day]
+}
+
+async function getTasksFromNote() {
```
nmn, line 44: `async function getTasksFromNote(): Promise<Array<TParagraph>> {`

weyert.ReminderHelpers/src/reminders.js
```diff
@@ -0,0 +1,368 @@
...
+  toISODateString,
+  smartPrependPara,
+} from '../../helperFunctions'
+
+// ------------------------------------------------------------------------------------
+const DEFAULT_REMINDERS_CONFIG = {}
+
+const minimumConfig = {}
+
+function getDateFromString(dateStr: string): Array<int> {
```
nmn, line 26: `Array<number>`
Javascript doesn't have int. Please make sure you have flow running locally.

nmn (changes requested):
There are a couple of things that need to change, and probably some testing you could do. Other than that it looks good enough.

weyert.ReminderHelpers/src/reminders.js
weyert: Oh okay, I will do :D

dwertheimer (changes requested):
@weyert This is looking really good. There are Flow errors in reminders.js that should be resolved before the merge. Most of them are benign. Many have to do with console.logs and ensuring you cast things in the console.log as String() or if they could be undefined, I like to say `console.log(```${foo || ''}```)`. If you run Flow locally in your IDE (e.g. https://github.com/NotePlan/plugins/tree/strftime#visual-studio-code-recommended) you will see all the errors and can fix them. I still have a few Flow errors floating around in my code -- hoping @nmn can have a look one day. But we're trying to do the best we can with Flow typing so that when people include your modules into their code, they work properly.
Screen Shot 2021-08-14 at 6 51 37 AM

dwertheimer:
Also, the README says that the plugin requires configuration. But the DEFAULT_REMINDERS_CONFIG and minimumConfig are empty. Unless this is just a placeholder for future work you want to do, it would be better to get rid of the configuration steps altogether and just have the plugin "configured" through the template tag. The _configuration note is a placeholder until proper preferences are available in the app. I generally try to avoid it unless the user gets some important benefit from it.

mikeerickson:
We should revisit this plugin after the changes to NotePlan plugin settings have been implemented.  I like the idea of this plugin, we just need to wait at this point as configuration changes will be different in the very near future.

@weyert @dwertheimer Please review accordingly when the new API has been released.

dwertheimer:
Is quite stale now and has merge conflicts. I will close until @weyert wants to update.

weyert:
Cool, yes, going to revisit it next week :)

PR #50, opened 2021-08-14
add `formatted` command to DateAutomation

mikeerickson:
# About PR
Added new command to DateAutomation

**IMPORTANT: Install `strftime` node_module before testing locally**

```bash
npm install strftime
```

*****

## Details

`formatted` will display date/time using `format` key `_configuraton` if exists, otherwise will use the default format `%Y-%m-%d %I:%M:%S %P`

## Configuration Changes

`format` key will exist in any `_configuration` templates starting with v0.9.0 (pending further changes before new version is posted)

New default `_configuration.date` block

```javascript
  ...
  date: {
    // Default timezone for date and time.
    timezone: 'automatic',
    // Default locale to format date and time.
    // e.g. en-US will result in mm/dd/yyyy, while en_GB will be dd/mm/yyyy
    locale: 'en-US',
    // can be "short", "medium", "long" or "full"
    dateStyle: 'medium',
    // optional key, can be "short", "medium", "long" or "full"
    timeStyle: 'short',
    // optional custom format (uses strftime format)
    // see https://www.strfti.me/ to aid in creating custom formats)
    format: '%Y-%m-%d %I:%M:%S %P'
  },
  ...
```


dwertheimer (approved):
Looks great. I only wonder whether we should move the underlying formatting function to helperFunctions so it can be used by other plugins.

PR #51, opened 2021-08-16
Add format and fix weather description

mikeerickson:
- Added `date.format` property (will be used by upcoming DateAutomations `formatted` feature)
- Updated `weather.openWeatherUnits` comment use `imperial` for Fahrenheit


dwertheimer:
@mikeerickson Thx for that fix. I assume we're waiting on this until there's a strftime? So do you just want to add these changes to the other [WIP] PR?

PR #65, opened 2021-08-29
Jgclark/issue57

jgclark:
Readying QuickCapture v0.7.0.
But has some knock-on changes to other plugins, as I've needed to tweak configuration code.
I know this is about to cause merge conflicts with my own code from earlier today. (Developed on two different Macs -- long story ...)


jgclark:
Also does EventHelpers v0.4.0

Issue #74, opened 2021-09-04
(low priority) pickDate and pickDateInterval - would be nice if they would run without params in template mode


dwertheimer:
@jgclark I have a Templates test script and I test all the templates after I touch that code.
The new pickDate and pickDateInterval doesn't run perfectly without params. Would be good if it provided feedback.

![Screen Shot 2021-09-03 at 5 26 57 PM](https://user-images.githubusercontent.com/8949588/132078878-7be17ffe-a344-48da-8720-11ef555c5053.jpg)

```js
# Template Tester
---

## Today:
8601 date the hard way:
- {{date({locale: 'sv-SE', dateStyle: 'short'})}}
Formatted:
- {{formattedDateTime('%Y-%m-%d %I:%M:%S %P')}}
8601 the easy way:
- {{date8601()}}
pickdate:
- {{pickDate()}}
- {{pickDateInterval()}}
week span:
- {{weekDates()}}
- {{weekDates({startMonday:false, format:`%Y-%m-%d`})}}
- {{weekDates({startMonday:true, format:`%Y-%m-%d`})}}
meetingName tag:
- {{meetingName}}
weather:
- {{weather()}} 
quote:
- {{quote()}} 
Tasks last 7 days with headings (with confirmation):
- {{sweepTasks({limit:{ "unit": "day", "num":1 },includeHeadings:true,requireConfirmation:true})}}
---
Events:
{{listTodaysEvents({template:"- *|START|*-*|END|*: *|TITLE|*",allday_template:"- *|TITLE|*"})}} ---
Matching events:
{{listMatchingEvents()}} -- (requires configuration)
```

jgclark:
@dwertheimer, I have now provided default questions for each. Please review PR to see if that does what you hope. 
(I'm not wanting to make it smarter that this, as I hope it will be replaced soon by new API calls from Eduard.)

jgclark:
It's been a week, so closing this one.

PR #75, opened 2021-09-04
Address #74 by adding default questions

jgclark:
@dwerheimer please review proposed solution


dwertheimer:
@jgclark This looks a lot better. Is the output supposed to be just what I typed in?
For instance, if I call it with no params, it asks me for a date, and whatever I type, it just inserts that in the text
same for "22w" or whatever. doesn't do any calculation. maybe that's correct. just asking what I should be expecting.

jgclark:
> @jgclark This looks a lot better. Is the output supposed to be just what I typed in?
> For instance, if I call it with no params, it asks me for a date, and whatever I type, it just inserts that in the text
> same for "22w" or whatever. doesn't do any calculation. maybe that's correct. just asking what I should be expecting.

Yup, nothing clever here.
For now they're more as placeholders for native API calls/UI that I hope Eduard will add.
OK, I'll commit (resolving the conflicts that have emerged since I created the PR).

dwertheimer:
Great!

PR #76, opened 2021-09-05
/qtn and {{getWeekDates()}}

dwertheimer:
OK @mikeerickson, I created a nice little PR here with some details about the new /qtn command and the {{getWeekDates()}} template tag. Normally, I would open this PR, and immediately merge it, because who's got time to do code review? What do you want the PR merge process to look like for us since we are all limited in time? Should I leave this PR here until you see and merge it? 
cc: @jgclark just for process visibility (you don't need to review)

### DIFF
--- a/Plugin-Listing.md
+++ b/Plugin-Listing.md
@@ -62,4 +62,5 @@
 | `tst` | Tasks Sort by #Tag - Sort tasks on the page by hashtag | ✅ Task Automations |
 | `tt` | Tasks to Top - Bring all tasks in note to top | ✅ Task Automations |
 | `wc` | Shows word count and other numbers for the note | 🔢 Note Statistics |
+| `wd` | Insert dates of current week | 📅 Date Automations |
 | `🔌 Plugin Tester` | Simple plugin test to check that plugins are working for you | 🔌 Plugin Tester |
--- a/dwertheimer.DateAutomations/src/dateFunctions.js
+++ b/dwertheimer.DateAutomations/src/dateFunctions.js
@@ -240,17 +240,16 @@ export async function formattedDateTimeTemplate(paramStr: string = ''): Promise<
 }
 
 //TODO FIXME: figure out formats and locales - WIP because the startMondy doesn't work
-// {startMonday:true, format:`'EEE yyyy-MM-dd'} // see [date-fns format](https://date-fns.org/v2.23.0/docs/format)
+// {weekStartsOn:1, format:`'EEE yyyy-MM-dd'} // see [date-fns format](https://date-fns.org/v2.23.0/docs/format)
 export async function getWeekDates(paramStr: string = ''): Promise<string> {
-  const startMonday = await getTagParamsFromString(paramStr, 'startMonday', true)
+  const weekStartsOn = await getTagParamsFromString(paramStr, 'weekStartsOn', 1)
   const format = await getTagParamsFromString(paramStr, 'format', 'EEE yyyy-MM-dd')
   console.log(startOfWeek(new Date(), { weekStartsOn: 0 }))
   console.log(startOfWeek(new Date(), { weekStartsOn: 1 }))
-  console.log(startMonday)
-  console.log(dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: startMonday ? 1 : 0 })), 'yyyy-MM-dd'))
-  const start = dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: startMonday ? 1 : 0 })), 'EEE yyyy-MM-dd')
+  console.log(dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: weekStartsOn })), 'yyyy-MM-dd'))
+  const start = dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: weekStartsOn })), 'EEE yyyy-MM-dd')
   console.log(start)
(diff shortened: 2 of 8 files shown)


mikeerickson:
I think this approach is a great start. At our current stage, not all features or bug fixes need to be PRs, but will get us in the workflow of review those items which need a formal process 

we can all use our discretion as to when a PR might be necessary. Sometimes, the timeline dictates how it is released. 

If we are doing bug fixes, then a PR approach might not be the best course of action, as it may be critical to end user feedback 

Having said all of that, give me a couple hours and I will have time to escape to the code dungeon (a rare feat on Sunday) and I will get back to you 👍

nmn.Templates/README.md
```diff
@@ -109,5 +110,28 @@ The `*|TITLE|*`, `*|START|*` and `*|END|*` can be mixed with whatever markdown c
 
 You can also place  `{{matchingEvents()}}` or `{{listMatchingEvents()}}` in Templates in a similar way, and similar customisation is possible. However, it is defined in a different way, using the matches and template strings defined in the \_configuration file's `addMatchingEvents` array, as shown above.
 
+### weekDates - ouput the beginning and end span/dates of the current week (e.g. Mon-Sun)
```
mikeerickson, line 113: I see a lot of information in the `nmn.Templates/README.md` but there is no associated `/qtn` in this PR so not sure what this is references.

nmn.Templates/README.md
```diff
@@ -109,5 +110,28 @@ The `*|TITLE|*`, `*|START|*` and `*|END|*` can be mixed with whatever markdown c
...
+-  `{{weekDates({weekStartsOn:1, format:"EEE yyyy-MM-dd"})}}` -- get the span of this week 
+	- e.g. `Mon 2021-08-30 - Sun 2021-09-05` (mon-sun by default) 
+	- weekStartsOn parameter is 1 (Monday) by default. Change it to 0 (Sunday) or another day to start the weeks on a different day
+	- see [date-fns format](https://date-fns.org/v2.23.0/docs/format) for formatting details.
+
+## `/qtn - Quick Template Note`
```
mikeerickson, line 119: I found this in the `nmn.Templates/plugin.json` but there is no short command, thus it won't execute inline or via command bar).  I would suggest adding `qtn` as an alias (or maybe I am using wrong as I have never used the `nmn.Templates` plugin

```json
    {
      "name": "qtn - Quick template note",
      "alias": ["quick", "template"],
      "description": "Choose a note and create it based on _configuration settings quickNotes",
      "jsFunction": "quickTemplateNote"
    } 
```

Plugin-Listing.md
```diff
@@ -62,4 +62,5 @@
 | `tst` | Tasks Sort by #Tag - Sort tasks on the page by hashtag | ✅ Task Automations |
 | `tt` | Tasks to Top - Bring all tasks in note to top | ✅ Task Automations |
 | `wc` | Shows word count and other numbers for the note | 🔢 Note Statistics |
+| `wd` | Insert dates of current week | 📅 Date Automations |
```
mikeerickson, line 65: I believe `qtn` needs to be added to `nmn.Templates/plugin.json` (which will then be reflected in this list.

dwertheimer.DateAutomations/src/dateFunctions.js
```diff
@@ -240,18 +240,14 @@ export async function formattedDateTimeTemplate(paramStr: string = ''): Promise<
...
-  console.log(startOfWeek(new Date(), { weekStartsOn: 0 }))
-  console.log(startOfWeek(new Date(), { weekStartsOn: 1 }))
-  console.log(startMonday)
-  console.log(dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: startMonday ? 1 : 0 })), 'yyyy-MM-dd'))
-  const start = dateFormat(new Date(startOfWeek(new Date(), { weekStartsOn: startMonday ? 1 : 0 })), 'EEE yyyy-MM-dd')
-  console.log(start)
-  const end = dateFormat(new Date(endOfWeek(new Date(), { weekStartsOn: startMonday ? 1 : 0 })), 'EEE yyyy-MM-dd')
-  console.log(end)
+  const weekStartsOn = Number(await getTagParamsFromString(paramStr, 'weekStartsOn', 1))
+  const format = String(await getTagParamsFromString(paramStr, 'format', 'EEE yyyy-MM-dd'))
```
mikeerickson, line 246: Which section of the `_configuration` note is this added?  Since I have never used this plugin, I am sure my `_configuration` file is bare.  However, should this not be using the `nmn.Templates/src/configuration/getOrMakeConfigurationSection` function?

mikeerickson:
OK, I have a reviewed the PR and have a few general comments:

1. make sure to create a new version number `x.x.x-pr.1` so that it is obvious when reviewing plugin list in `NotePlan Preferences > Plugins`
2. make sure to specify what command(s) you wish to be further tested in description (I am not sure what is new or what was added/fixed) so that one can get there as quickly as possible.  
3. I tried to run `/qtn` but the code for the PR does not contain this in the updated `nmn.Templates` plugin (it is referenced in the `nmn.Templates/README.md` file
4. Further, in reviewing the PR code, I see an addition to `Plugin_Listing.md` so I tried that command
5. That command works (inserting `Mon 2021-08-30 - Sun 2021-09-05` at the cursor)

OK, that is all for now, I believe I have tested everything (but still curious about `/qtn` reference)

mikeerickson (approved):

PR #77, opened 2021-09-07
Changing `---` in Templates from a requirement to a nice-to-have

dwertheimer:
@jgclark @nmn The initial work on Templates was before my time, but I see the hard requirement for a horizontal rule. The first time I created a template, I forgot the `---` and didn't understand why it didn't work (until I re-read the README). Now we've heard from at least 2 users who have been tripped up by it. I am suggesting in this PR that we make the <HR> a nice-to-have rather than a hard requirement. If you have it, great. If not, we'll just use the whole note (other than the title). Unless there's some reason the <HR> needs to be a hard requirement? (in which case we should add better error reporting). Thoughts?

### DIFF
--- a/nmn.Templates/changelog.md
+++ b/nmn.Templates/changelog.md
@@ -2,6 +2,8 @@ See Template Plugin [README](https://github.com/NotePlan/plugins/blob/main/nmn.T
 
 # What's Changed in this Plugin?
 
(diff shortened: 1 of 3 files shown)


jgclark:
Good idea, and sensible-looking implementation.
I can't test it right now, but I think you need to check the minimal case of a template with just a heading and one more line. There's a line with [string, string, string] in it ...

PR #78, opened 2021-09-07
Fix: /todayStart wasn't running if you didn't have a calendar day open

dwertheimer:
fix: fixed an error in /todayStart that kept it from running if the note wasn't open. It now opens it automatically.
Not sure how I missed this before, but fixed it. Bumped version so it could be released whenever you get a chance to look @jgclark


jgclark (approved):

dwertheimer:
Thanks @jgclark!

PR #80, opened 2021-09-09
Fix config setup problem

jgclark:
Reported 8.9.2021 by @MrCoffee and @colingold
Someone added a single newline in template-folder.js without realising the knock-on impact.


mikeerickson:
@jgclark So, in reading your comment here (and what you posted o plugin channel) I am guessing the offending “new line” was in `templates-folder.js` line 154?   Pretty crazy that such a small, seemingly innocent change could cause so much harm.

jgclark:
Yes, I think it was. Just a result of the way the
insert-new-section-into-first-codeblock happened to be implemented. Neither
wrong on their own but together they were …--
Regards, -- Jonathan

mikeerickson:
Man, I bet that was fun tracking down.  One of those things that “looks” correct to the human eye, but the computer eye says something different.  Regardless, great job nonetheless.

PR #106, opened 2021-11-17
Feature/favorites

dwertheimer:
@mikeerickson This was a small plugin. Could have been about 15 total lines of code. I think I wrote about 150, so there could be nice clean tests and separation between pure code and Noteplan code. I'm quite happy with how it turned out, but I'm sure it can be even better. If you have suggestions for improvements, let me know.


mikeerickson:
I will circle back on this one after `np.Templating` and new CLI projects are out the door.   Not closing so that I can review again later.

PR #110, opened 2021-11-21
add .idea folder to gitignore file

m1well:
* added `.idea` folder to gitignore file for opening the code in IntelliJ
* deleted `.pluginpath` entry, because it is duplicated


dwertheimer:
Thanks for noticing that duplicate

PR #113, opened 2021-11-22
refactor: lint error curly braces and orderd import and changed one import to chalk

m1well:
as discussed earlier - a separate pull request for "refactor: lint error curly braces and orderd import and changed one import to chalk"


mikeerickson:
I am going to close this one as `rollup.js` has been modified quite a bit in the new CLI project, thus I am going to defer to those changes.  We can circle back on this one at a later date.

dwertheimer:
This reminds me @codedungeon that the .pluginpath is set in the autowatch (and presumably the CLI). We should probably include our newfound Setapp path in there and stat for the existence of the different paths to set it automatically, and if they both exist (like they will for @jgclark and @eduard), use one of your CLI ask-a-question thingies.

PR #114, opened 2021-11-22
refactor: simplified if-else

m1well:
as discussed earlier - a separate pull request for "refactor: simplified if-else"


mikeerickson:
This is actually a placeholder method as more validation is forthcoming so I am going to close out this one.  But, thanks for input :-)

PR #116, opened 2021-11-22
fix: errors with optional parameters

m1well:
i fixed two errors with optional parameters (I wrote about that in the 'plugin-dev' channel)

`func(paramString?: string = '')` - this is double optional

either one should do `func(paramString: string = '')` -> then if there is no parameter it is set to an empty string
or one should do `func(paramString?)` -> this is cleaner, BUT then you need nullchecks in the following code

so I decided to go with the first option, just to get rid of this errors.

also I commented something in the `stats.js` because my IDE throws a lot of errors there.

### DIFF
--- a/jgclark.EventHelpers/src/eventsToNotes.js
+++ b/jgclark.EventHelpers/src/eventsToNotes.js
@@ -122,7 +122,7 @@ export async function getEventsForDay(
 
 //------------------------------------------------------------------------------
 // Return MD list of today's events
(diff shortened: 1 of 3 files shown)


mikeerickson:
I am going to defer to @jgclark for his review on this one as he is more familiar with inner workings and can advise quicker

jgclark:
Accepting and committing.
(The src/stats.js issues are known; they're from WIP that escaped to main. Hopefully I can revert that to stop errors appearing for others.)

PR #117, opened 2021-11-22
refactor: eslint added unused imports plugin

m1well:
just if you say it is ok - then merge ;)

I found so many unused imports, that I thought I add the plugin to ESLint - then one can find out easily if there are unused imports.
(by the way there are still some other ESLint errors...)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,8 @@
   ],
   "plugins": [
     "import",
-    "flowtype"
+    "flowtype",
+    "unused-imports"
   ],
   "parser": "@babel/eslint-parser",
   "parserOptions": {
@@ -56,7 +57,9 @@
     "no-prototype-builtins": "off",
     "no-unused-vars": "off",
     "import/order": "error",
-    "no-useless-escape": "off"
+    "no-useless-escape": "off",
+    "unused-imports/no-unused-imports": "error",
+    "unused-imports/no-unused-vars": "off"
   },
   "env": {
     "node": true,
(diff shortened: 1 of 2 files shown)


mikeerickson:
@m1well I will get this reviewed shortly!

dwertheimer:
@mikeerickson What do you think about merging this one? I probably have lots of these ghost imports in my code -- I don't often go back and look at imports after refactoring. (note to self). Can we pull this one into the repo (assuming it works for us in VSCode -- I know @m1well does it in IntelliJ)?

jgclark:
Unless my memory is faulty, the eslint settings some time back did report unused imports, at least for me in VSCode.  Clearly this has changed, and I'd not noticed the unnecessary imports since.

So I'm happy for you to clean up my files in this way.

dwertheimer:
Given how much time has passed, I would like to suggest a step-wise approach: @m1well reduces this PR to just the .eslint settings, we merge and then everyone cleans up their own plugins as they edit.

I think the Rollup tree-shaking deletes any errant imports, so I don't think it matters much and so there is no urgency IMO, but it will clean up the code a lot.

m1well:
> Given how much time has passed, I would like to suggest a step-wise approach: @m1well reduces this PR to just the .eslint settings, we merge and then everyone cleans up their own plugins as they edit.
> 
> I think the Rollup tree-shaking deletes any errant imports, so I don't think it matters much and so there is no urgency IMO, but it will clean up the code for the future.

ok I updated the PR.

dwertheimer:
Thanks @m1well!

PR #119, opened 2021-11-24
fixed userinput number integer log and added leftpad method

m1well:


jgclark (approved):
Thanks for the additions and corrections.

PR #125, opened 2021-11-29
affirmations added flow and custom configured affirmations

m1well:
* another idea from me: get a random custom configured affirmation. because then one could configure some affirmations in another language and get a random one with the tag `{{affirmation(true)}}`
* also I added `@flow` to the `affirmations.js` for better type safety

### DIFF
--- a/helpers/dev.js
+++ b/helpers/dev.js
@@ -44,3 +44,14 @@ export function logAllPropertyNames(obj) {
   console.log(Object.getOwnPropertyNames(obj).filter((x) => /^__/.test(x) === false))
   logAllPropertyNames(Object.getPrototypeOf(obj))
 }
+
+/**
+ * @description get a random element of an array
+ * @author m1well
+ *
+ * @param array array with n elements
+ * @returns {any} random element
+ */
+export const getRandomElementFromArray = (array) => {
+  return array[Math.floor((Math.random() * array.length))]
+}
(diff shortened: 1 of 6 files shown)


dwertheimer:
@m1well This is great. I like the feature additions and the changes you made to make the code more resilient. I think the challenge is that all the underlying code has been re-written in @mikeerickson's new Templates system (coming soon). So we could merge this now, but it will get wiped out when the Templates system is replaced. I'm thinking that maybe you wait on this until Templates 2.0 gets released and then drop this code into that system. All your code is good and modular, so it should be simple to just paste it into the right sections using Templates2. Thoughts?

m1well:
> @m1well This is great. I like the feature additions and the changes you made to make the code more resilient. I think the challenge is that all the underlying code has been re-written in @mikeerickson's new Templates system (coming soon). So we could merge this now, but it will get wiped out when the Templates system is replaced. I'm thinking that maybe you wait on this until Templates 2.0 gets released and then drop this code into that system. All your code is good and modular, so it should be simple to just paste it into the right sections using Templates2. Thoughts?

yes of course I can wait ;)
for me it would be nicer if the other PRs are getting merged.
this here was just an idea to make the affirmations more flexible. but I can change this PR after I know what I have to change for the new template system

jgclark:
I echo both parts of what David says. If it was any other week, I'd have committed the change and released it.
It should be obvious when the new Templating system is released.

mikeerickson:
As per @dwertheimer and @jgclark i will close this one as the new templating system has this and a lot more goodies around this overall API access.  Almost there game, almost there.

PR #126, opened 2021-11-29
[WIP] autotimeblocking - see notes

dwertheimer:
@jgclark U wanted a copy of the code to test for conflicts, so here you go
@mikeerickson I'm not actually suggesting you merge this PR into main, but I'd like to hear from you how I should deal with a large commit like this... 95% of the code is limited to my plugins, but I did add some helpers etc. I was thinking maybe I should cherry-pick the helper code and do a separate PR for those. But pls scan through this PR and give me some direction so I can do it correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "eslint-plugin-import": "^2.23.2",
     "flow-bin": "^0.151.0",
     "inquirer": "^8.1.0",
-    "jest": "27.0.6",
+    "jest": "^27.3.1",
     "mkdirp": "^1.0.4",
     "prettier": "^2.3.0",
     "rollup": "^2.48.0",
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,7 +9,8 @@
   ],
   "plugins": [
     "import",
-    "flowtype"
+    "flowtype",
+    "unused-imports"
   ],
   "parser": "@babel/eslint-parser",
   "parserOptions": {
@@ -56,7 +57,9 @@
     "no-prototype-builtins": "off",
     "no-unused-vars": "off",
     "import/order": "error",
-    "no-useless-escape": "off"
+    "no-useless-escape": "off",
+    "unused-imports/no-unused-imports": "error",
+    "unused-imports/no-unused-vars": "off"
   },
   "env": {
     "node": true,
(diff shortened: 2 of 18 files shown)


mikeerickson:
When cross pollinating (fancy word) code form other libraries, they should indeed be in a commit all to themselves.  However, for the sake of getting your plugin out to the masses to start the testing loop, I say just do it and I will deal with merge conflicts on my end when I unleash templating.  I don’t want to hold things up for stuff like this.

mikeerickson:
@dwertheimer OK, after reviewing this whole commit, I have some comments that you might want to consider before pushing this puppy live.  I have not looked at code at all, just surface-level things to consider.

1. Since you are modifying code across multiple plugins, it is now more than just library code that could cause an issue.  Since you modified `dwertheimer.EventAutomations` and modified `dwertheimer.TaskAutomations` you will have to create beta builds for both of them to satisfy your beta requirements and dependencies.

2. With the changes and additions to `helpers`, you are running the risk of breaking changes for any other plugins which might be using these helpers.  I really doubt that is gonna happen, but just something to think about it.

3. This reminds me that I have been thinking about adding a `package.json` to the `helpers` folder so that it can be version tracked like global `package.json` and individual `plugin.json` is doing for the rest of the project

4. Speaking of potential breaking changes for existing users, I don't see a change in `plugin.json` for `dwertheimer.TaskAutomations` which should certainly exist since you have modified the plugin?

5. Thanks for adding tests for the new and modified `helpers`!

6. Did the `jest` change in `package.json` address a bug you might of run into with jest, or was this merely getting jest to the latest version.   Since you jumped `minor` release, there is a chance that it will introduce breaking changes to other jest installations (mine) but I *REALLY* doubt that and can confirm this by updating my local version of jest and testing all my plugins, etc.

What did you have planned for the case of item 1 and item 2? Will the changes made to `dwertheimer.TaskAutomations/src/taskHelpers.js` cause any breaking changes for existing users of those plugins?

dwertheimer:
Thanks for the feedback. Exactly what I was looking for.

package.json
```diff
@@ -60,7 +60,7 @@
     "eslint-plugin-import": "^2.23.2",
     "flow-bin": "^0.151.0",
     "inquirer": "^8.1.0",
-    "jest": "27.0.6",
```
dwertheimer, line 63: You are correct that this was just an npm update thing...not anything I did on purpose. So we can ignore it.

PR #130, opened 2021-12-01
eventhelpers eventlist show calendarname

m1well:
* since I do not know which files are all affected on the new templating system from @mikeerickson, the only thing I can do is to bring up my ideas - if you don't merge them then it's a pity but ok
* because I also see the calendar of my wife, it would be nice to have the possibility to see the calendar name in the eventslist
* also I added a mapping, so that you can map the calendar name, maybe for formatting purposes (as it is for me)

### DIFF
--- a/jgclark.EventHelpers/CHANGELOG.md
+++ b/jgclark.EventHelpers/CHANGELOG.md
@@ -4,7 +4,12 @@ See [website README for more details](https://github.com/NotePlan/plugins/tree/m
 - tighten timeblock-finding regex if Eduard does
 - add further fields (e.g. location) if Eduard adds to the API
 
-## [0.7.0] - 2021-19-11
+## [0.8.0] - 2021-12-01 (@m1well)
+### Added
+- added boolean variable to config to render calendar name in eventslist
+- added possibility to map a calendar name to a given string (maybe for formatting purposes)
+
+## [0.7.0] - 2021-11-19
 ### Added
 - added `/process date offsets` command: find date offset patterns and turn them into due dates, based on date at start of section, or a less-indented line, or the line itself.
 
--- a/jgclark.EventHelpers/README.md
+++ b/jgclark.EventHelpers/README.md
@@ -37,20 +37,24 @@ Or add the following settings into the `Templates/_configuration` note's first c
 ```jsonc
 ...
   events: {
-    calendarToWriteTo: "", // specify calendar name to write events to. Must be writable calendar. If empty, then the default system calendar will be used.
+    calendarToWriteTo: "",  // specify calendar name to write events to. Must be writable calendar. If empty, then the default system calendar will be used.
     addEventID: false,  // whether to add an '⏰event:ID' string when creating an event from a time block
-    processedTagName: "#event_created",   // optional tag to add after making a time block an event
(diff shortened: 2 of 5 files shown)


mikeerickson:
All these changes will have no impact on what is being done in the new templating plugin, thus they are safe to change.  I am going to leave this to @jgclark for his review as he knows the inner workings of this plugin.

jgclark (commented):
Two comments on this:
1. Why have you introduced the `pref_showCalendarName`, as I can't see a purpose for it? Surely we can leave it to whether the template includes `*|CAL|*` or not?

2. Several times in the templates and settings you have `...*|CAL|**|TITLE|*...`. Won't these two need a space between them?

m1well:
> Two comments on this:
> 
> 1. Why have you introduced the `pref_showCalendarName`, as I can't see a purpose for it? Surely we can leave it to whether the template includes `*|CAL|*` or not?
> 2. Several times in the templates and settings you have `...*|CAL|**|TITLE|*...`. Won't these two need a space between them?

to 1.:
it is used in the method `getReplacements` where all String "regex" are getting replaced:
 `value: pref_showCalendarName ? calendarNameWithMapping(item.calendar, pref_calendarNameMappings) : ''`
 
 to 2.:
yes I wanted to have a space between `CAL` and `TITLE` - but the problem is: because it is the first element.  
if you don't want to see the calendar name, then you have an empty space in front of your text.
example (white spaces are underlines to have a better look)
normal line:
`* hello_world`
line with CAL:
`* calname:_hello_world`
without cal:
`* _hello_world`

jgclark:
> to 1.: it is used in the method `getReplacements` where all String "regex" are getting replaced: `value: pref_showCalendarName ? calendarNameWithMapping(item.calendar, pref_calendarNameMappings) : ''`

I understand how you've used it, but I was asking why you need it at all. It looks to me like you don't need this setting at all.
 
> to 2.: yes I wanted to have a space between `CAL` and `TITLE` - but the problem is: because it is the first element. if you don't want to see the calendar name, then you have an empty space in front of your text. 

Ah, yes, I see. Perhaps one way around this is once the whole line is constructed is to then run a replace that replaces any consecutive whitespace characters with a single space. I.e. \\h{2,} → \\s

m1well:
> > to 1.: it is used in the method `getReplacements` where all String "regex" are getting replaced: `value: pref_showCalendarName ? calendarNameWithMapping(item.calendar, pref_calendarNameMappings) : ''`
> 
> I understand how you've used it, but I was asking _why you need it_. It still looks to me like you don't need this setting at all.
> 
> > to 2.: yes I wanted to have a space between `CAL` and `TITLE` - but the problem is: because it is the first element. if you don't want to see the calendar name, then you have an empty space in front of your text.
> 
> Ah, yes, I see. Perhaps one way around this is once the whole line is constructed is to then run a replace that replaces any consecutive whitespace characters with a single space. I.e. \h{2,} → \s


now I understood -> you are right - this boolean value is not needed because one could handle this via the `|CAL|` in the template string. I removed it.
also I added this replacement of multiple spaces with one.
but now because this boolean value is no longer needed - there is no case for this two spaces. but I added it anyway ;)

jgclark (approved):
I can see a line that needs removing still in timeblocks.js but I'm going to approve and sort it out from here to save time.

PR #133, opened 2021-12-06
trim user input - and use it in expenses and journal plugin

m1well:
* new function for trimmed user input
* used now in expenses plugin and journal plugin


jgclark (approved):
All good from reading the code.

PR #134, opened 2021-12-07
dailyjournal todaystart configurable templatename

m1well:
* template name for `/todayStart` and `/dayStart` is now configurable 
* i also added a hint in the README that one has to be careful with `/dayStart` on another day than today with tags like `date... / formattedDate...` and `weather()`

### DIFF
--- a/jgclark.DailyJournal/src/journal.js
+++ b/jgclark.DailyJournal/src/journal.js
@@ -2,7 +2,7 @@
 //--------------------------------------------------------------------------------------------------------------------
 // Daily Journal plugin for NotePlan
 // Jonathan Clark
-// last update v0.9.0, 25.11.2021 by @jgclark/@m1well
+// last update v0.10.0, 07.12.2021 by @jgclark/@m1well
 //--------------------------------------------------------------------------------------------------------------------
 
 import { isInt, showMessage } from '../../helpers/userInput'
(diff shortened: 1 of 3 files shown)


jgclark.DailyJournal/src/journal.js
```diff
@@ -2,7 +2,7 @@
 //--------------------------------------------------------------------------------------------------------------------
 // Daily Journal plugin for NotePlan
 // Jonathan Clark
-// last update v0.9.0, 25.11.2021 by @jgclark/@m1well
+// last update v0.10.0, 07.12.2021 by @jgclark/@m1well
```
jgclark, line 5: Last update is just you not me as well!

jgclark.DailyJournal/src/journal.js
```diff
@@ -54,6 +56,29 @@ export async function dayStart(today: boolean = false): Promise<void> {
...
+  {
+    console.log('\tWarning: Cannot find suitable \'dailyJournal\' settings in Templates/_configuration note. Stopping.')
+    await showMessage(
+      'Cannot find \'dailyJournal\' settings in _configuration.',
+      'Yes, I\'ll check my _configuration settings.',
+    )
+    return
+  }
+
+  pref_templateTitle = (journalConfig?.nameOfDailyNoteTemplate != null)
```
jgclark, line 78: Using two very different names for the same thing is confusing.
Can I suggest you use `templateTitle` not `nameOfDailyNoteTemplate` which better fits with the style of other settings, and matches the `pref_templateTitle` in the code?

jgclark (changes requested):
A helpful addition, thanks.
I'm just requesting a couple of stylistic changes. And an update to the CHANGELOG.

m1well:
done @jgclark

jgclark (approved):

PR #136, opened 2021-12-07
weather api mulitple language support

m1well:
* added multiple language support to the weather api

### DIFF
--- a/nmn.Templates/README.md
+++ b/nmn.Templates/README.md
@@ -45,15 +45,15 @@ Weather: 🌧️ Moderate rain 14/19°C
 - `{{date({locale: 'sv-SE', dateStyle: 'short'})}}` -- Date borrowing the Swedish "Locale" yields ISO-8601 date like `2021-06-21`
 - `{{date8601()}}` -- ISO-8601 date like `2021-06-21` (shorter tag, no options)
 - `{{formattedDateTime({format: '%Y-%m-%d %I:%M:%S %P'})}}` -- date/time now using [strftime](https://www.strfti.me/) format (pass the version you want)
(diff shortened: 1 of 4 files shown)


mikeerickson:
@m1well While there is a good chance this PR will be closed, it is not a lost cause as two things have been exposed in accordance with this PR

1. Before taking the time and energy to modify the code and produce PR, it is always a good idea to first submit a RFF which allows you to share your thoughts as to what problem you are wanting to solve and your desired implementation
2. It has exposed the fact that a better overall method for language support is needed

As @dwertheimer mentioned on Discord https://discord.com/channels/763107030223290449/809886801857871883/917804241295396924 for `i18n` support, this is definitely something that needs to advance in both NotePlan and plugin implementation.

I have mentioned here https://discord.com/channels/763107030223290449/809886801857871883/917807393730101299 that the upcoming `np.Templating` plugin uses system language setting for anything that would need language translation, and the same approach should be the preferred language interface for all future plugin development

mikeerickson:
Note:  This PR will remain open as a reminder that further development will be necessary for language support, but actual implementation will be different based on other language implementation for NotePlan plugins as a whole.

jgclark:
@mikeerickson I agree with the need for a more planned and systemic approach to i18n.
However, I'd be inclined to implement this PR, as it is taking advantage of a pre-built i18n option in the 3rd party API that this plugin calls. It therefore devolves most of the job *in this particular case* to the place which has worked this through already, and will be able to do a better job behind its scenes that we would.

mikeerickson:
Closing this PR as core plugin API will be different as it relates to string localization. More information will be provided when the API and NotePlan have been updated to support user language translation

PR #141, opened 2021-12-09
expenses plugin redesign

m1well:
after a discussion with @dwertheimer I created a whole redesign of my expenses plugin.
some of the changes:
* columorder is configurable
* delimiter is configurable
* 3 more commands for faster usage
* more exception handling
* more tests
* more type checks on some areas
* ...

### DIFF
--- a/helpers/general.js
+++ b/helpers/general.js
@@ -169,15 +169,3 @@ export function capitalize(s: string): string {
   if (typeof s !== 'string') return ''
   return s.charAt(0).toUpperCase() + s.slice(1)
 }
-
-/**
- * @description here you can left pad your number with zeros - e.g. a '5' with 3 targetDigits is getting a '005'
- * @author m1well
- *
- * @param current the current number
- * @param targetDigits how many digits should the target number have
- * @returns {string} the left padded value as string
- */
-export const leftPadWithZeros = (current: number, targetDigits: number): string => {
-  return String(Array(Math.max(targetDigits - String(current).length + 1, 0)).join('0') + current)
-}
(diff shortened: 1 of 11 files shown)


helpers/general.js
```diff
@@ -169,15 +169,3 @@ export function capitalize(s: string): string {
   if (typeof s !== 'string') return ''
   return s.charAt(0).toUpperCase() + s.slice(1)
 }
-
-/**
```
dwertheimer, line 173: did you mean to move this out of helpers/general? normally you would want to do this as a separate PR since this is a global helper file. Probably ok because nobody else seems to have used it since you added it. But why did you move it?

.gitignore
```diff
@@ -17,3 +17,4 @@ package-lock.json
 lcov*
 coverage/
 .history/*.*
+xtemp/
```
dwertheimer, line 20: what's xtemp?

dwertheimer (approved):
Looks really great. I tested it and works! Thanks!

.gitignore
m1well: just a folder where you can put some files which aren't tracked via git.

e.g.:
* sometimes I just copy my "real expenses" to my IDE to have a backup if I am testing in Noteplan.
* or I had to rewrite all `2021;02` dates to `2021-02` and I did that in my IDE

helpers/general.js
m1well: unfortunately i had to move this function because my tests did not work anymore.
in this `general` there is a reference to `DataStore` and then the test always breaks, because I don't know how to mock the `DataStore`.
Of course I don't move functions which are in use by anybody else ;)

.gitignore
dwertheimer: feels like everyone else doesn't need this though. can you do this locally and discard the change from the PR?

helpers/general.js
dwertheimer: got it. makes sense.

.gitignore
m1well: ok :D

PR #146, opened 2021-12-15
quickcapture flow fix and reformatting

m1well:
quickcapture plugin
* fixed flow error
* did some code reformatting


mikeerickson:
@jgclark It looks like most @m1well proposed changes in `src` files is attributed to ESLint rules requiring singe quotes vs double quotes.

jgclark (approved):

PR #147, opened 2021-12-15
expenses plugin amount format

m1well:
added possibility to choose
* `full` amount format (always with 2 fraction digits)
* or `short` as it currently is - with no fraction digits and always rounded


dwertheimer:
Looks great @m1well. Thanks so much. I tested in en-US format (. separator) and it works perfectly.

jgclark (approved):
I've done a visual check through only.
I have just one suggestion: in `expensesChecks.js` I find this description doesn't make sense (in English) `* @param currentDate current date - to have always the same`. I think from your code I might have said "example of date string to check format is valid". 

But as that's a suggestion only, I will approve and merge for you.

PR #148, opened 2021-12-15
expenses.js small refactoring

m1well:
* changes from @jgclark

### DIFF
--- a/m1well.Expenses/src/expensesChecks.js
+++ b/m1well.Expenses/src/expensesChecks.js
@@ -34,8 +34,8 @@ export const categoryOk = (category: string, categories: string[]): boolean => {
  * @description just do some checks on the privided config and e.g. add a default delimiter if none is set
  *
  * @param config casted config from _configuration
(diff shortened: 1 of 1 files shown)


dwertheimer:
My suggestion for little changes like this would be to change them in your code and then just have them be part of the next substantial commit

m1well:
> My suggestion for little changes like this would be to change them in your code and then just have them be part of the next substantial commit

but why?
1. the plugin isn't released in the newest version
2. it doesn't affect real code, so there won't be a need for a new release (if it was released - what it is not)
3. I don't know when I am going to do a next commit for the expenses plugin :D

dwertheimer:
To answer your question, I think it's noisy to create pulls for things like simple JSDoc changes. Yes, it can be done (and I will do this time), but everyone is notified unnecessarily, and it creates more git history for non-essential items. I tend to keep these types of changes in my working branch and commit them down the road when I have something meaningful.

PR #149, opened 2021-12-17
initial version

nhueck:
plugin for Noteplan: nhueck.groupTasks 
sorts and groups open Tasks according to their hashtags, clears empty lines between them

### DIFF
--- a/nhueck.GroupTasks/plugin.json
+++ b/nhueck.GroupTasks/plugin.json
@@ -1,0 +1,17 @@
+{
+  "noteplan.minAppVersion": "3.0.21", 
+  "plugin.id": "nhueck.GroupTasks", 
+  "plugin.name": "🖇GroupTasks", 
+  "plugin.description": "Group tasks by their hashtag in current note", 
+  "plugin.author": "Nikolaus Hueck",
+  "plugin.version": "0.0.4", 
+  "plugin.dependencies": ["nmn.Templates"], 
+  "plugin.script": "script.js",
+  "plugin.commands": [ 
+	{
+	  "name": "groupOpenTasks", 
+	  "description": "Group open tasks by their hashtag in current note", 
+	  "jsFunction": "groupTasks"
+	}
+  ]
+}
--- a/nhueck.GroupTasks/src/groupTasks.js
+++ b/nhueck.GroupTasks/src/groupTasks.js
@@ -1,0 +1,241 @@
+
+//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
+// Noteplan Plugin groupTasks
+// groups all open tasks of the current note according to their first #tag
+// and adds them below a heading "Tag"
+// Author: Nikolaus Hueck twitter: @nikolaushueck
+//------------------------------------------------------------------------------
+//------------------------------------------------------------------------------
+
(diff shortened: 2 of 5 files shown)


mikeerickson:
Hello @nhueck 

In reviewing your initial PR, I am curious:

- Are your desires to release this as a public NotePlan Plugin, or will this be for internal use only

I can do a more thorough review based on knowledge of your desired use case.

nhueck.GroupTasks/src/groupTasks.js
```diff
@@ -0,0 +1,241 @@
...
+  return s.replace('#', '')
+}
+
+//---------------------------------------
+// helper function: stripAsteriskFromTodo(s)
+// Deletes all "*{space}"s in a string
+//---------------------------------------
+
+function stripAsteriskFromTodo(s) {
+  return s.replace('* ', '')
```
jgclark, line 88: I don't think this is a good way of doing this, as
a) there are at least two other ways of marking tasks.
I suggest you instead look at the type ParagraphType ("open" | "done" | "scheduled" | "cancelled" | "title" | "quote" | "list" | "empty" | "text") to look for "open" task types.

nhueck.GroupTasks/src/groupTasks.js
```diff
@@ -0,0 +1,241 @@
...
+//---------------------------------------
+// helper function: deleteHeading(text,level)
+// Delete a heading a given text and level
+//---------------------------------------
+function deleteHeading(text,level) {
+  const levelstring = "#".repeat(level)
+  console.log("deleting title: "+text+" level: "+ level)
+  const replace = '(\n)*?'+levelstring+' '+text+' *?(\n)*?'
+  const re = new RegExp(replace,"g")
+  Editor.content = Editor.content.replace(re,'')
```
jgclark, line 122: I think this will replace all occurrences of the heading in the note. Will this always be what you want. If so, I suggest you need to update the documentation (and function title) to make it clear.

nhueck.GroupTasks/plugin.json
```diff
@@ -0,0 +1,17 @@
+{
+  "noteplan.minAppVersion": "3.0.21", 
+  "plugin.id": "nhueck.GroupTasks", 
+  "plugin.name": "🖇GroupTasks",
```
jgclark, line 4: Needs a reference to the readme.md file

jgclark (commented):
As well as the individual comments here are a couple of more general ones:
- in the last couple of months we've started to use jsdoc style for documentation in the js files.  Please see most of the recently-edited files for examples of this.
- after the first version, we have a CHANGELOG.md file as well.

I haven't checked the logic of the main function.

nhueck.GroupTasks/src/groupTasks.js
```diff
@@ -0,0 +1,241 @@
...
+  deleteEmptyTasksHeadings: true
+}
+
+//---------------------------------------
+// helper function: getPriority(string)
+// get the priority og a given task
+// returns a number (1 - 3) 3 beeing the highest priority
+//---------------------------------------
+function getPriority(s) {
+
```
jgclark, line 35: Presumably this isn't complete.

PR #150, opened 2021-12-20
helper flexible answers to one question via input

m1well:
* flexible amount of answers to one question via userinput
* also define minimum and maximum amount of answers


dwertheimer:
Looks great. Squashed and merged. Assuming @m1well tested edge cases.

PR #152, opened 2021-12-20
refactor: eslint versions - we cant use v8

m1well:
sorry for the circumstances - unfortunately we cannot use the version like this :(


mikeerickson:
having to revert to ESLint v7 as v8 implementation is causing issues elsewhere.

PR #155, opened 2021-12-26
helper to clear note

m1well:
* helper to clear note


jgclark (commented):
This will be useful to have, and reminds me that I have the equivalent of a removeSection function that I could move to the helpers.
But anyway, I'm sure this works, though I'd have tried using the `removeParagraphs` API call instead. The only thing I suggest you check is whether doing the paragraph-by-paragraph removal lands up creating many more versions than if a single `removeParagraphs` call is made?

jgclark (approved):
Approving after code read-through only.

PR #156, opened 2021-12-26
listdaysevents sort if with calendarname

m1well:
* I added this `|CAL|` option for the calendar name some days ago
* now I added the sorting of this calendar names (and also the start times)


jgclark (approved):
Thanks for these improvements. 
I've only done a code read through now, but can approve, merge, and then release later on today, hopefully, after some real testing.

PR #158, opened 2021-12-26
introduce SimpleParagraph for better paragraph processing

m1well:
* introduce SimpleParagraph for better paragraph processing
* example create new paragraphs:
```
const lines: SimpleParagraph[] = []
...
lines.push({content: 'hello world', type: 'list'})
...
lines.forEach(line => note.appendParagraph(line.content, line.type))
```
* example mapping from existing paragraph
```
const lines: SimpleParagraph[] = []
note.paragraphs.forEach(paragraphToSimpleParagraphMapper())
...
```

### DIFF
--- a/helpers/paragraph.js
+++ b/helpers/paragraph.js
@@ -30,6 +30,7 @@ export function parasToText(paras: Array<TParagraph>): string {
 /**
  * Print out all data for a paragraph as JSON-style string
  * @author @EduardMe
+ *
  * @param {TParagraph} p - paragraph to print
  */
 export function printParagraph(p: TParagraph) {
@@ -79,11 +80,12 @@ export function printParagraph(p: TParagraph) {
  * Works out which line to insert at top of file. Rather than just after title line,
  * go after any YAML frontmatter or a metadata line (= starts with a hashtag).
  * @author @jgclark
+ *
  * @param {TNote} note - the note of interest
  * @return {number} line - the calculated line to insert/prepend at
  */
 export function calcSmartPrependPoint(note: TNote): number {
-  const lines = note.content?.split('\n') ?? ['']
+  const lines = note.content?.split('\n') ?? [ '' ]
 
   // By default we prepend at line 1, i.e. right after the Title line
   let insertionLine = note.type === 'Calendar' ? 0 : 1
@@ -134,6 +136,7 @@ export function calcSmartPrependPoint(note: TNote): number {
  * I.e. if the note starts with YAML frontmatter (e.g. https://docs.zettlr.com/en/core/yaml-frontmatter/)
  * or a metadata line (= starts with a hashtag), then add after that.
  * @author @jgclark
+ *
  * @param {TNote} note - the note to prepend to
  * @param {string} paraText - the text to prepend
  * @param {ParagraphType} paragraphType - the usual paragraph type to prepend
@@ -147,11 +150,11 @@ export function smartPrependPara(
   note.insertParagraph(paraText, calcSmartPrependPoint(note), paragraphType)
 }
 
-//
 /**
  * Works out where the first ## Done or ## Cancelled section starts, if present.
  * If not, return the last paragraph index.
  * @author @jgclark
+ *
  * @param {TNote} note - the note to assess
  * @return {number} - the index number
  */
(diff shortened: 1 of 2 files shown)


jgclark:
I'm not wanting to be difficult, but I don't see how this helps. From your first example it seems that we still need to remember to specify .type separately from .content throughout. 

I think either helper functions (or better, API extensions) to separate real line content from the other bits and pieces (date markers, task markers) would be more helpful.

helpers/paragraph.js
```diff
@@ -181,4 +181,33 @@ export function findEndOfActivePartOfNote(note: TNote): number {
...
+}
+
+/**
+ * @description Function to map a TParagraph to a SimpleParagraph, to only have content and type
+ * @example note.paragraphs.map(paragraphToSimpleParagraphMapper(())
+ * @author @m1well
+ *
+ * @returns {function(TParagraph): {type: ParagraphType, content: string}|null}
+ */
+export const paragraphToSimpleParagraphMapper = (): Function => {
```
dwertheimer, line 193: This is interesting. I'm curious why this function returns a function. Why did this need to be a composition?

dwertheimer (commented):
This all looks great code-wise. I'm just not sure I understand the use case which drove the need for this new type. Can you describe what need you had? I can tell you that a few months ago I had started to develop a function which added additional metadata to paragraphs (e.g. priority, etc.) in a regular JS object which could be console.logged (one of my frustrations with NP developing). At the time, some of the .paragraph fields were not reliable (indents, heading), there was no .children field, etc. so I was calculating them all. Maybe these ideas should be combined. Here's what my [WIP] code looked like (the two different times I started to do something like this):

image

image

I realize that what I'm doing is not Simple at all, so it's probably a totally different use case, but thought we should discuss.

helpers/paragraph.js
m1well: because then you can use it directliy in the `map` part:
```
yesterdayNote.paragraphs
          .filter(para => para.type === 'open')
          .map(paragraphToSimpleParagraphMapper()) <----- here
          .map(simpleParagraphScheduledFromYesterday()) <----- this is the same principle
```

also the same principle like my sorting I suggested also in the plugin-dev channel:
```
export const sortByPrio = (): Function => {
  return (second: TParagraph, first: TParagraph) => {
    if (haveSameTypes(first, second)) {
      if (isFlaggedThrice(first) || isFlaggedThrice(second)) {
        return isFlaggedThrice(first) ? 1 : -1
      }
      if (isFlaggedTwice(first)) {
        return isFlaggedThrice(second) ? -1 : 1
      }
      if (isFlaggedTwice(second)) {
        return isFlaggedThrice(first) ? 1 : -1
      }
      if (isFlaggedOnce(first)) {
        return (isFlaggedTwice(second) || isFlaggedThrice(second)) ? -1 : 1
      }
      if (isFlaggedOnce(second)) {
        return (isFlaggedTwice(second) || isFlaggedThrice(first)) ? 1 : -1
      }
    }
    return 0
  }
}
```
call:
```
doneTasks
            .sort(sortByPrio()) <---- here
            .map(cleanupParagraphs())
```

so this can avoid to always write functions which has an array as input parameter, do something, and then return the changed array ;)

dwertheimer:
Ah that’s cool

m1well:
@jgclark  @dwertheimer 
I think you are "on the wrong way":
this PR is not for the purpose to split information from the `content`part (as I wrote in my ideas/wishes for the api in CANNY).
I can try to explain it again. given following code where I use it.
i have this array for "questions about yesterday" - like 
```
      yesterdayQuestions: [
        'What did you eat',
        'What were you grateful for',
      ],
```
and I also use my function for multiple answers I published some days ago (userInput.js:368).
and that leads to following code:
```
      // init an array for the new lines in a daily note
      const lines: SimpleParagraph[] = []
      
      // .... then do some other stuff
      
      // then the first part to show how `SimpleParagraph` works - "cache" lines in the array
      for (let i = 0; i < config.yesterdayQuestions.length; i++) {
        const answers = await multipleInputAnswersAsArray(config.yesterdayQuestions[i], 'OK / Leave empty to finish answers', true, 1, 2)
        if (answers.length > 0) {
          lines.push({ content: `### ${config.yesterdayQuestions[i]}`, type: 'text' }) <---- HERE!!
          answers.forEach(ans => lines.push({ content: ans, type: 'list' }))
        }
      }
      
      // then at the end, write all cached lines in the note AT ONE TIME
      for (let i = start; i < lines.length + start; i++) {
        note.insertParagraph(lines[i].content, i, lines[i].type)
      }
```

question: HOW do you create kind of a "cache" for paragraphs?
solution:
* to insert or append a paragraph to a note, what do you need? CONTENT and TYPE
* so I created this "simple" type which holds CONTENT and TYPE ;)

otherwise:
* how do you create a cache like this:
```
      const lines: TParagraph[] = []
```
and then push lines to it??? I guess you have to push all the things here:
```
      lines.push({ content: `### ${config.yesterdayQuestions[i]}`, type: 'text', rawContent: '', prefix: '', lineIndex: number, date: new Date(), heading: '', ........... }) 
```

hopefully this explains my need for that type :)

m1well:
> In the example, is the (()) correct?
> Couldn't you simply pass the name of the function?
> note.paragraphs.map(paragraphToSimpleParagraphMapper)

I thought that, too. but no, for whatever reason i get "undefined" as return in the respective line.
don't know why...

dwertheimer:
@m1well I think that if the function were just the inside part of what you wrote, you could just pass the function to map without invoking it (the parentheses). I have not tested it, but I think this should work:

const paragraphToSimpleParagraphMapper = (paragraph) => paragraph.content ? {
      content: paragraph.content,
      type: paragraph.type,
    } : emptySimpleParagraph()
  }

const simpleParas = notes.paragraphs.map(paragraphToSimpleParagraphMapper)

m1well:
> @m1well I think that if the function were just the inside part of what you wrote, you could just pass the function to map without invoking it (the parentheses). I have not tested it, but I think this should work:
> 
> const paragraphToSimpleParagraphMapper = (paragraph) => paragraph.content ? { content: paragraph.content, type: paragraph.type, } : emptySimpleParagraph() }
> 
> const simpleParas = notes.paragraphs.map(paragraphToSimpleParagraphMapper)

unfortunately this does not work with flow  :(

PR #159, opened 2021-12-28
folders test fix

m1well:


jgclark:
That's a great help, @m1well! Thanks for sorting out what I'd missed in the tests, and for further improvements in the functions themselves.

PR #161, opened 2021-12-28
template selection

m1well:
* add selection to templates


mikeerickson:
I don't have a problem merging this in for now, but it will need to be implemented a bit differently in the new Templating plugin, but I will worry about those details.

@dwertheimer Do you have time to review the PR locally and make sure it all works?  On the surface, seems it will be just fine.  If all works as expected, feel free to release.

@m1well Thanks for the temporary implementation :-)

jgclark:
This will be a good addition. (Assuming the selection is still available via `Editor.selectedParagraphs` at this stage in the process of applying a template.)

dwertheimer:
Thanks @m1well

PR #163, opened 2021-12-30
weather api fix

m1well:
see discussion in Discord "using-noteplan"
* config null check
* README

### DIFF
--- a/nmn.Templates/CHANGELOG.md
+++ b/nmn.Templates/CHANGELOG.md
@@ -2,6 +2,10 @@ See Template Plugin [README](https://github.com/NotePlan/plugins/blob/main/nmn.T
 
 # What's Changed in this Plugin?
 
+## [0.14.0] - 2021-12-31 (@m1well)
+### Changed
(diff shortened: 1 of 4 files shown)


jgclark (changes requested):
Not approved:
- we don't yet know what the problem is for this user
- according to SemVer this sort of change would be 0.13.1
- I'd appreciate the " not being changed to ', particularly then when ' need to be escaped to work properly

The only issue discovered so far is problem in the README having quotes around the lat/long fields, which I have addressed.

m1well:
> Not approved:
> 
> * we don't yet know what the problem is for this user
> * according to SemVer this sort of change would be 0.13.1
> * I'd appreciate the " not being changed to ', particularly then when ' need to be escaped to work properly
> 
> The only issue discovered so far is problem in the README having quotes around the lat/long fields, which I have addressed.

I accept that with the version (0.13.1), everything else I can not understand at all. the error is obvious, I even tested it.
but it is ok.

PR #164, opened 2021-12-31
Fix/tt insertion index bug

dwertheimer:


dwertheimer.TaskAutomations/src/sortTasks.js
```diff
@@ -177,7 +177,11 @@ function insertTodos(note: TNote, todos, heading = '', separator = '', subHeadin
     .join(`\n`)
   console.log(`Inserting tasks into Editor:\n${contentStr}`)
   // console.log(`inserting tasks: \n${JSON.stringify(todosWithSubheadings)}`)
-  note.insertParagraph(`${headingStr}${contentStr}${separator ? `\n${separator}` : ''}`, 1, 'text')
+  note.insertParagraph(
+    `${headingStr}${contentStr}${separator ? `\n${separator}` : ''}`,
+    note.type === 'Calendar' ? 0 : 1,
+    'text',
+  )
 }
```
dwertheimer, line 185: @jgclark You may want to look at this change. The bug you found (thank you) was due to the fact that the TOP index of the content of a Project Note is 1 but the top index of a Calendar note is zero because there's no title. I'm gonna make a PR for you to look at.

PR #165, opened 2021-12-31
Update insertionIndex code to deal with any type of note

dwertheimer:
git shows lots of changes (I guess due to Lint config), but I only made one change (line 89)

### DIFF
--- a/helpers/paragraph.js
+++ b/helpers/paragraph.js
@@ -86,7 +86,7 @@ export function calcSmartPrependPoint(note: TNote): number {
   const lines = note.content?.split('\n') ?? ['']
 
   // By default we prepend at line 1, i.e. right after the Title line
-  let insertionLine = 1
+  let insertionLine = note.type === 'Calendar' ? 0 : 1
   // If we have any content, check for these special cases
   if (lines.length > 0) {
     if (lines[0] === '---') {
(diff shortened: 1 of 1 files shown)


helpers/paragraph.js
```diff
@@ -142,16 +138,12 @@ export function calcSmartPrependPoint(note: TNote): number {
  * @param {string} paraText - the text to prepend
  * @param {ParagraphType} paragraphType - the usual paragraph type to prepend
  */
-export function smartPrependPara(
-  note: TNote,
-  paraText: string,
-  paragraphType: ParagraphType,
-): void {
+export function smartPrependPara(note: TNote, paraText: string, paragraphType: ParagraphType): void {
```
jgclark, line 141: When a func definition has more than a couple of parameters, following nmn's lead, I lay them out this way to make it easier to read.

helpers/paragraph.js
```diff
@@ -173,12 +165,7 @@ export function findEndOfActivePartOfNote(note: TNote): number {
       cancelledHeaderLine = i
     }
   }
-  const endOfActive =
-    (doneHeaderLine > 0)
-      ? doneHeaderLine
-      : ((cancelledHeaderLine > 0)
-        ? cancelledHeaderLine
-        : lineCount)
+  const endOfActive = doneHeaderLine > 0 ? doneHeaderLine : cancelledHeaderLine > 0 ? cancelledHeaderLine : lineCount
```
jgclark, line 168: I can't read nested ? : without layout structure.
I don't mind you removing brackets but please leave the layout.

jgclark (changes requested):
Thanks for finding this, @dwertheimer.
Approved, bar the couple of layout comments below.

helpers/paragraph.js
dwertheimer: TBH, I didn't (intentionally) make these formatting changes, VSCode/Prettier/.eslint must have reformatted on save. I don't know how to tell it not to. Do you or @mikeerickson? Are you using the standard .eslintrc settings in the repo?

jgclark: Aha.
I appear to have prettier turned off. I guess I got fed up of it changing things like this where occasionally I find particular layouts helpful.
I think I'm using the standard .eslintrc file.

jgclark:
I've made the logic mods but left the layout as was.

PR #174, opened 2022-01-04
updated the plugin-skeleton

m1well:
* updated plugin.json
* changed hello world function to new arrow-function-style

### DIFF
--- a/np.plugin-flow-skeleton/src/helloWorld.js
+++ b/np.plugin-flow-skeleton/src/helloWorld.js
@@ -4,6 +4,7 @@
 // export default async function [name of the function called by Noteplan]
 // Type checking reference: https://flow.org/
 // Specific how-to re: Noteplan: https://github.com/NotePlan/plugins/blob/main/Flow_Guide.md
+
 export async function helloWorld(): Promise<void> {
   const message = 'Hello World from {{pluginName}}!'
 
--- a/np.plugin-flow-skeleton/CHANGELOG.md
+++ b/np.plugin-flow-skeleton/CHANGELOG.md
@@ -15,7 +15,7 @@ All NotePlan plugins follow `semver` versioning. For details, please refer to [s
 
 See Plugin [README](https://github.com/NotePlan/plugins/blob/main/{{pluginId}}/README.md) for details on available commands and use case.
 
-## [x.x.x] - yyyy-mm-dd (githubUserName)
+## [x.x.x] - yyyy-MM-dd (@githubUserName)
 
 ### Added
 List what has been added. If nothing has been changed, this section can be removed.
--- a/np.plugin-flow-skeleton/plugin.json
+++ b/np.plugin-flow-skeleton/plugin.json
@@ -1,15 +1,16 @@
 {
   "COMMENT1": "Note If you are not going to use the `npm run autowatch` command to compile, then delete the macOS.minVersion line below",
-  "macOS.minVersion": "10.13.0",
   "noteplan.minAppVersion": "3.0.21",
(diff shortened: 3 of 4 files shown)


np.plugin-flow-skeleton/src/helloWorld.js
```diff
@@ -4,7 +4,8 @@
 // export default async function [name of the function called by Noteplan]
 // Type checking reference: https://flow.org/
 // Specific how-to re: Noteplan: https://github.com/NotePlan/plugins/blob/main/Flow_Guide.md
-export async function helloWorld(): Promise<void> {
+
+export const helloWorld = async (): Promise<void> => {
```
mikeerickson, line 8: This will be the only piece I would reject as most all other plugins in this repository use the original signature `export async`, changing to `export const` would introduce more friction.  

This can be left to the plugin developer as to the method they are most comfortable with.

mikeerickson:
@m1well If you would update the PR to address the items mentioned in file comments, I would be happy to accept and merge this PR.

np.plugin-flow-skeleton/src/helloWorld.js
m1well: ok

m1well:
> @m1well If you would update the PR to address the items mentioned in file comments, I would be happy to accept and merge this PR.

I force pushed the changes.
ATTENTION: I also removed this old generated script.js from the folder

np.plugin-flow-skeleton/src/helloWorld.js
jgclark: Not a comment, but a JS question: what is the benefit of the `export const ... async ...` style? I would certainly not fully understand this style.

mikeerickson: @jgclark have a look at this article, it breaks it down quite well.

[Variable Declaration vs Function Express](https://medium.com/@mandeep1012/function-declarations-vs-function-expressions-b43646042052)

FWIW, I would often side with using variable declaration, vs function expression, but like I mentioned to @m1well almost all of the plugins, helpers, et. Use function declaration, thus I was more look for this item to be consistent with what we have.

@m1well In general, I too would use variable declaration, but again, wanting to stay consistent :-)

mikeerickson (approved):
Approving and merging this PR now so that it can get into Plugins master, but want to let you know that this entire template will be removed in the next major release of Plugin API (version 3) which will come with CLI v3.

CLI v3 will bring move this template from top level, into the CLI assets

np.plugin-flow-skeleton/src/helloWorld.js
dwertheimer: At the risk of kicking off a religion debate...it's mostly stylistic and doesn't matter, but...IMO, it would be nice to have some kind of standardization in the codebase stylistically -- especially when we're sharing helper code. Always bugs me to see a mix of styles in one code file. 

I don't do this for a job like @mikeerickson or @m1well, so take this with a grain of salt, but I'm a huge fan of Airbnb opensourcing their lint rules to that end -- keep everyone writing the same style of code. I personally tend to favor:
a) variable+anonymous functions for one-liners (e.g. no curly braces needed). If I need a curly brace, I turn it into a function declaration.
b) function declarations for all else
My experience in debugging/stack tracing is that function-name inference on variable functions is inconsistent, and it's easier to see how you got here through named functions.

Besides, it's way easier (for me) when scanning through code to see that something's a function versus someone setting a variable.

Sounds like "your mileage may vary" in opinions on this issue. So, let the style war comments begin!

PR #175, opened 2022-01-05
introduce new utils plugin

m1well:
PR because of this requirement to automatically archive specific notes `using-noteplan` channel.
i had something similar and collected now some of my useful functions I currently use nearly daily.

please read the README, this is too much for here :D
in short:
- `repeater` - repeat functionality with many possibilities and just by selecting any line(s) in every note you want
- `sorter` - sort lines by type (open, scheduled, cancelled, done, list, text) and prio (`!!!`, `!!`, `!`)
- `autoArchiver` - archive notes with a specific hashtag and a specific lifetime
- `cleanUpEmptyLinesInFuture` - just clean up empty lines in future daily notes

### DIFF
--- a/m1well.Utils/plugin.json
+++ b/m1well.Utils/plugin.json
@@ -18,10 +18,10 @@
       "jsFunction": "repeater"
     },
     {
-      "name": "sorter",
-      "alias": ["well:sorter", "well"],
+      "name": "selectionSorter",
+      "alias": ["well:selectionSorter", "well"],
       "description": "Sort selected lines by type and by prio",
-      "jsFunction": "sorter"
+      "jsFunction": "selectionSorter"
     },
     {
       "name": "archiveNotes",
--- a/m1well.Utils/plugin.json
+++ b/m1well.Utils/plugin.json
@@ -1,0 +1,39 @@
+{
+  "noteplan.minAppVersion": "3.0.25",
+  "macOS.minVersion": "10.13.0",
+  "iOS.minVersion": "14",
+  "plugin.id": "m1well.Utils",
+  "plugin.name": "🛠 Utils",
+  "plugin.description": "Plugin for some Utils like repeat selected task, etc.",
+  "plugin.author": "@m1well",
+  "plugin.url": "https://github.com/NotePlan/plugins/blob/main/m1well.Utils/README.md",
+  "plugin.version": "1.0.0",
+  "plugin.dependencies": [],
+  "plugin.script": "script.js",
+  "plugin.commands": [
+    {
+      "name": "repeater",
+      "alias": ["util:repeater", "util"],
+      "description": "Repeat task(s) just by selecting them in whatever note you want with a lot of functionalities",
+      "jsFunction": "repeater"
+    },
+    {
+      "name": "sorter",
+      "alias": ["util:sorter", "util"],
+      "description": "Sort selected lines by type and by prio",
+      "jsFunction": "sorter"
+    },
+    {
+      "name": "autoArchiveNotes",
+      "alias": ["util:autoArchiveNotes", "util"],
+      "description": "Provides possibility to auto archive notes with a specific tag and a given lifetime (moves note to @Archive)",
+      "jsFunction": "autoArchiveNotes"
+    },
+    {
+      "name": "cleanUpEmptyLinesInFuture",
+      "alias": ["util:cleanUpEmptyLinesInFuture", "util"],
+      "description": "Remove empty lines in future daily notes",
+      "jsFunction": "cleanUpEmptyLinesInFuture"
+    }
+  ]
+}
--- a/m1well.Utils/plugin.json
+++ b/m1well.Utils/plugin.json
@@ -13,25 +13,25 @@
   "plugin.commands": [
     {
       "name": "repeater",
-      "alias": ["util:repeater", "util"],
+      "alias": ["well:repeater", "well"],
       "description": "Repeat task(s) just by selecting them in whatever note you want with a lot of functionalities",
       "jsFunction": "repeater"
     },
     {
       "name": "sorter",
-      "alias": ["util:sorter", "util"],
+      "alias": ["well:sorter", "well"],
       "description": "Sort selected lines by type and by prio",
       "jsFunction": "sorter"
     },
     {
-      "name": "autoArchiveNotes",
-      "alias": ["util:autoArchiveNotes", "util"],
-      "description": "Provides possibility to auto archive notes with a specific tag and a given lifetime (moves note to @Archive)",
-      "jsFunction": "autoArchiveNotes"
+      "name": "archiveNotes",
+      "alias": ["well:archiveNotes", "well"],
+      "description": "Provides possibility to archive notes with a specific tag and a given lifetime (moves note to @Archive)",
+      "jsFunction": "archiveNotes"
     },
     {
       "name": "cleanUpEmptyLinesInFuture",
-      "alias": ["util:cleanUpEmptyLinesInFuture", "util"],
+      "alias": ["well:cleanUpEmptyLinesInFuture", "well"],
       "description": "Remove empty lines in future daily notes",
       "jsFunction": "cleanUpEmptyLinesInFuture"
     }
--- a/m1well.Utils/src/cleanup.js
+++ b/m1well.Utils/src/cleanup.js
@@ -14,6 +14,8 @@ import {
   sortByType
 } from './utilsHelper'
 
+export const TASK_LIST_TEXT_TYPES = ['open', 'scheduled', 'cancelled', 'done', 'list', 'text']
+
 const CONFIG_KEYS = {
   archiveNotesTag: 'archiveNotesTag',
   archiveNotesLifeInDays: 'archiveNotesLifeInDays',
@@ -39,7 +41,7 @@ const EXAMPLE_CONFIG = `
  *
  * @private
  */
-const sorter = async (): Promise<boolean> => {
+const selectionSorter = async (): Promise<boolean> => {
 
   const selectedParagraphs = Editor.selectedParagraphs
 
@@ -55,11 +57,8 @@ const sorter = async (): Promise<boolean> => {
       }
     }
 
-    const selectedNonTasksAndList = selectedParagraphs.filter(para => para.type !== 'open' && para.type !== 'scheduled'
-      && para.type !== 'done' && para.type !== 'cancelled' && para.type !== 'list')
-
-    const selected = selectedParagraphs.filter(para => para.type === 'open' || para.type === 'scheduled'
-      || para.type === 'done' || para.type === 'cancelled' || para.type === 'list')
+    const selectedNonTasksAndList = selectedParagraphs.filter(para => !TASK_LIST_TEXT_TYPES.includes(para.type))
+    const selected = selectedParagraphs.filter(para => TASK_LIST_TEXT_TYPES.includes(para.type))
 
     Editor.removeParagraphs(selectedParagraphs)
 
@@ -174,4 +173,4 @@ const validateConfig = (config: Config): string | null => {
   return null
 }
 
-export { sorter, archiveNotes, cleanUpEmptyLinesInFuture }
+export { selectionSorter, archiveNotes, cleanUpEmptyLinesInFuture }
--- a/m1well.Utils/src/cleanup.js
+++ b/m1well.Utils/src/cleanup.js
@@ -1,0 +1,177 @@
+// @flow
+
+import { addDays, differenceInCalendarDays, startOfTomorrow } from 'date-fns'
+import { getOrMakeConfigurationSection } from '../../nmn.Templates/src/configuration'
+import type { Config } from './utilsHelper'
+import {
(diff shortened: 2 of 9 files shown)


jgclark:
Some general comments on this, from just looking at the README.
Most are about where the most logical place is to discover and find these utilities.  And I don't think it's by something called 'Utils' when there are other more specific options available.

### `/util:repeater`
This is 95% overlapping with the existing jgclark.RepeatExtensions. Please can you explain why we need both?

### `/util:sorter`
This is somewhat overlapping with what @dwertheimer has written. That needs a re-write ... Can this be a contribution to that work?

### `/util:autoArchiver`
We seem to use the verb 'archive' in two different ways: one to archive tasks within a note, and the other to archive whole notes.  I suggest you make it clearer by referring to 'autoArchiveNotes' in code and documentation.

Also, this feels like it belongs in the existing NoteHelpers plugin.  I'd be happy to have it moved there.

### `/util:cleanUpEmptyLinesInFuture`
This feels like it belongs in the existing NoteHelpers plugin.  I'd be happy to have it moved there.

m1well:
> 



> Some general comments on this, from just looking at the README. Most are about where the most logical place is to discover and find these utilities. And I don't think it's by something called 'Utils' when there are other more specific options available.
> 
> ### `/util:repeater`
> This is 95% overlapping with the existing jgclark.RepeatExtensions. Please can you explain why we need both?
> 
> ### `/util:sorter`
> This is somewhat overlapping with what @dwertheimer has written. That needs a re-write ... Can this be a contribution to that work?
> 
> ### `/util:autoArchiver`
> We seem to use the verb 'archive' in two different ways: one to archive tasks within a note, and the other to archive whole notes. I suggest you make it clearer by referring to 'autoArchiveNotes' in code and documentation.
> 
> Also, this feels like it belongs in the existing NoteHelpers plugin. I'd be happy to have it moved there.
> 
> ### `/util:cleanUpEmptyLinesInFuture`
> This feels like it belongs in the existing NoteHelpers plugin. I'd be happy to have it moved there.

I just wrapped up some of my utils and made them accessible via a command.
the reasons for this new plugin are following:
- for me it is difficult to read all that old code
- and it is even harder to adjust that code, especially when there are some additional code errors or flow-type errors
- also some of my ideas / PRs felt like a "fight" against "the core team", so that is why I just added a complete new plugin, to not change old code and/or affect some other plugins with my changes
- just wanted to share some - imho - useful functions
- and my repeater does not overlap 95% with the other one - it is a different concept - on the other one you need a completed task and this `@done....` - on my repeater you just need any line in any note -> so this would be hard to merge
- yes I can change this `autoArchiver` thing to `autoArchiveNotes` :)

mikeerickson:
I am going to interject here and address specific comments from @jgclark  and response from @m1well 

> for me it is difficult to read all that old code

I don’t think this is a viable reason to not PR against existing code.  My recommendation would be to work with overlapping plugin owner , to come to a solution to make existing plugin command better.

> and it is even harder to adjust that code, especially when there are some additional code errors or flow-type errors

Similar to above, I feel it would be better to “fix” these errors (do these errors prevent command from working?) and flow-type errors is a big mystery with all code that uses flow.   One day, it would be nice to address flow issues across the entire project.

> also some of my ideas / PRs felt like a "fight" against "the core team", so that is why I just added a complete new plugin, to not change old code and/or affect some other plugins with my changes

I can surely address this one, as I feel it is most directed at me as I am the one who rejected many of your early PRs.  They were rejected for many reasons, some of which were due the PRs against a large number of source files (address formatting or spelling errors etc) because many of them are already fixed in Plugin API v3 which will be released when it is ready.

The “the core team” comment is factual.  We have a core team for real reasons, and is not going to change anytime soon.  

> just wanted to share some - imho - useful functions

I have ZERO problems with this, and I would encourage this PR address the issues @jgclark mentioned and cleanup the README accordingly.  Put some “polish” on it and rename plugin to something like “Workflow Utilities”  If it has a clean plugin name and useful methods, by all means submit it.  Further, creating something like this which can be iterated on as time passes, by all means, submit an revised PR.

With regards to the namespace, I would suggestion something like “well:repeater” or “well:archiveNotes” (the word “auto” to mean says it does something automatically (with you user having to invoke command) which we know is not possible at this time with NotePlan plugins.  All commands require end user invoking. 

Further, with regards the repeater function, i think this is fine if it does something different.  Much like VSCode plugins, there are often overlapping commands as each plugin has it’s own way of doing things.

> as I also wrote at the start of the README, for me this would be OK to merge this in the future if we have a "api style" paradigm/plugin mechanism

I am not sure what you are implying here, could you clarify what this means as it relates to current plugin API?

Overall, we would be more than happy to see another new plugin, and you have added some functionality that is currently not available, and offering your own take on some other functions.  If you want to make the recommended adjustments, I would love to see them,

m1well:
this is what i called a "fight" - and i did not only mean mike ;)
i can't answer all the comments here.
i can do some renamings as suggested with this "well:....." and i can also throw out the "auto" on the "autoArchiveNotes".
i can only offer you my developer help and have already put a few hours into all this code and have another 1,000 lines of code locally.
but ultimately it is your decision whether you merge it or not...

@mikeerickson depending the "api style" david and i discussed about that in the `plugin-dev` channel on the day you said that you are not disinterested, but sick - that is why you had not participated on the discussion. i guess ist was 02.01./03.01.

m1well.Utils/src/cleanup.js
```diff
@@ -0,0 +1,177 @@
+// @flow
```
dwertheimer, line 1: Not a big deal but "cleanup" seems like a strange filename for sorter to be in. I wouldn't think to look there for task sorting code. So I had to trace it from index.

m1well.Utils/src/cleanup.js
```diff
@@ -0,0 +1,177 @@
...
+    // check at least 2 paragraphs against all paragraphs and then get the first line index
+    // otherwise it gets inserted at lineIndex = 0
+    for (let i = 0; i < Editor.paragraphs.length; i++) {
+      if (Editor.paragraphs[i].rawContent === selectedParagraphs[0].rawContent
+        && Editor.paragraphs[i + 1].rawContent === selectedParagraphs[1].rawContent) {
+        firstIndex = Editor.paragraphs[i].lineIndex
+      }
+    }
+
+    const selectedNonTasksAndList = selectedParagraphs.filter(para => para.type !== 'open' && para.type !== 'scheduled'
```
dwertheimer, line 58: This is fine, but for readability, I would prefer to see something like:
const TASK_TYPES = ["open","scheduled"...
and then in both .filter statements:
TASK_TYPES.indexOf(para.type) > -1

m1well.Utils/src/cleanup.js
```diff
@@ -0,0 +1,177 @@
...
+
+    const selected = selectedParagraphs.filter(para => para.type === 'open' || para.type === 'scheduled'
+      || para.type === 'done' || para.type === 'cancelled' || para.type === 'list')
+
+    Editor.removeParagraphs(selectedParagraphs)
+
+    selected
+      .sort(sortByType())
+      .sort(sortByPrio())
+      .forEach((para, i) => Editor.insertParagraph(para.content, firstIndex + i, para.type))
```
dwertheimer, line 69: this chaining is nice!

m1well.Utils/src/utilsHelper.js
```diff
@@ -0,0 +1,202 @@
...
+    return 0
+  }
+}
+
+/**
+ * sort by type (open, scheduled, cancelled, done, list, text) by it's type
+ *
+ * @returns {(function(TParagraph, TParagraph): (number))} sorted paragraphs
+ */
+export const sortByType = (): Function => {
```
dwertheimer, line 81: I find this super complicatated. I'm sure it works and you've tested it, but I am sure there's a shorter way to do this.

m1well.Utils/src/utilsHelper.js
```diff
@@ -0,0 +1,202 @@
...
+export const getDailyNote = (day: Date): ?TNote => {
+  return DataStore.calendarNoteByDate(day)
+}
+
+/**
+ * sort function for paragraphs by prio `!!!`, `!!` and `!`
+ *
+ * @returns {(function(TParagraph, TParagraph): (number))}
+ */
+export const sortByPrio = (): Function => {
```
dwertheimer, line 53: This is the only part of the code that really bugs me. Hard coding the flag levels seems limiting (only three levels?)  and unnecessary. You could have grabbed or used my code for this, which creates a numeric priority based on the arbitrary number of exclamation points. Some of the relevant code here:

const EXCLAMATIONS = /\B(!+\B)/g
const PARENS_PRIORITY = /^\s*\(([a-zA-z])\)\B/g // must be at start of content
export const TASK_TYPES = ['open', 'scheduled', 'done', 'cancelled']

function getElementsFromTask(content, reSearch) {
  const found = []
  let matches = reSearch.exec(content)

  do {
    if (matches?.length > 1) {
      found.push(matches[1].trim())
    }
  } while ((matches = reSearch.exec(content)) !== null)
  return found
}

/*
 * Get numeric priority level based on !!! or (B)
 */
function getNumericPriority(item) {
  let prio = -1
  if (item.exclamations[0]) {
    prio = item.exclamations[0].length
  } else if (item.parensPriority[0]) {
    prio = item.parensPriority[0].charCodeAt(0) - 'A'.charCodeAt(0) + 1
  } else {
    prio = -1
  }
  return prio
}

m1well.Utils/plugin.json
```diff
@@ -0,0 +1,39 @@
...
+  "plugin.script": "script.js",
+  "plugin.commands": [
+    {
+      "name": "repeater",
+      "alias": ["well:repeater", "well"],
+      "description": "Repeat task(s) just by selecting them in whatever note you want with a lot of functionalities",
+      "jsFunction": "repeater"
+    },
+    {
+      "name": "sorter",
```
dwertheimer, line 21: to not confuse users, this should be called "sort selected tasks" to be clear that this relates to selections only (and tasks)

dwertheimer (changes requested):
I only reviewed sorter. Most of these comments are stylistic and just in case you're interested. One substantial note about exclamations and one about the plugin name/alias.

m1well.Utils/plugin.json
m1well: it is not only for tasks ;)
but ok I change it to `selectionSorter` :)

m1well.Utils/src/utilsHelper.js
m1well: > This is the only part of the code that really bugs me. Hard coding the flag levels seems limiting (only three levels?) and unnecessary. You could have grabbed or used my code for this, which creates a numeric priority based on the arbitrary number of exclamation points. Some of the relevant code here:
> 
> const EXCLAMATIONS = /\B(!+\B)/g const PARENS_PRIORITY = /^\s*(([a-zA-z]))\B/g // must be at start of content export const TASK_TYPES = ['open', 'scheduled', 'done', 'cancelled']
> 
> function getElementsFromTask(content, reSearch) { const found = [] let matches = reSearch.exec(content)
> 
> do { if (matches?.length > 1) { found.push(matches[1].trim()) } } while ((matches = reSearch.exec(content)) !== null) return found }
> 
> /*
> 
> * Get numeric priority level based on !!! or (B)
>   */
>   function getNumericPriority(item) {
>   let prio = -1
>   if (item.exclamations[0]) {
>   prio = item.exclamations[0].length
>   } else if (item.parensPriority[0]) {
>   prio = item.parensPriority[0].charCodeAt(0) - 'A'.charCodeAt(0) + 1
>   } else {
>   prio = -1
>   }
>   return prio
>   }

from development view:
the point with the limitation is a good point -  limitation is not that good.
from theming view:
currently I guess there exists only `!!!`, `!!` and `!` (even I use only 2 levels) - so imho we can ignore more than 3 levels :D
but I can change this if you want - no problem.
I know you all are kind of "regex pros" - this is one of the most hated topics for real developer - so I am very bad in it. and for me it is hard to read those regexes. but I can take a look at yours and grab it from there

dwertheimer:
Or “sort selection”

m1well.Utils/src/utilsHelper.js
m1well: i don't know how - I just played around with these custom sort functions and this works as expected - there is even a test for it ;)

m1well.Utils/src/cleanup.js
m1well: because there are functions which "cleans up" a note ;)

m1well.Utils/src/cleanup.js
m1well: yes javascript is not my "mother language" :P
so i am learning the specific functions also day by day

m1well:
added now changes for  @dwertheimer .
only one point what is a bit "ugly" - this regex for the exclamation marks could be in the whole string not only at the beginning. so `normal task !!!! 1` would have a higher prio than `!!! hello world`.
I played around a bit with this regex but getting an exclamation mark at the beginning of a string with regex seems to be very hard (other than a `*`) ...

dwertheimer:
That’s by design. Exactly how I want it to work.

PR #176, opened 2022-01-05
expenses plugin - notes via datastore

m1well:
* my "# 1 expenses plugin user" @dwertheimer had the idea that the tracking notes should not open every time after tracking
* so I changed it to call the datastore instead 
* Also deleted my `@description` in the jsdoc - don't know how I could have forget them

### DIFF
--- a/m1well.Expenses/CHANGELOG.md
+++ b/m1well.Expenses/CHANGELOG.md
@@ -1,5 +1,9 @@
 # m1well.Expenses Plugin Changelog
 
+## [1.6.0] - 2022-01-05 (@m1well)
+### Changed
+- Load tracking notes via DataStore to not open them all the time
+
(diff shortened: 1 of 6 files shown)


mikeerickson:
@m1well I know it was discussed somewhere (likely #plugin-dev channel on discord) but I am curious why `@description` was removed from all your docBlocks?

m1well:
> @m1well I know it was discussed somewhere (likely #plugin-dev channel on discord) but I am curious why `@description` was removed from all your docBlocks?

Because it is not needed - then we decided to remove all the @description tags. But not sure why i forgot some of them...
It is only needed if you want to have the discription somewhere else but not on the top of the comment

dwertheimer:
Mike, that was my doing. I suggested that we try to always put the
description in the default location at the top and therefore not use the
description tag

PR #179, opened 2022-01-08
dev.js fix

m1well:


mikeerickson:
Why have you renamed the `environment` to `Environment` ?

m1well:
> Why have you renamed the `environment` to `Environment` ?  

Because it is `Environment.environment` if you look at flow-typed. I dont know why, that is not from me. But it is the same like `Editor.xxx` or `DataStore.xx`

mikeerickson:
Ah, I see what you mean there.   IMHO, I think it would be better to rename `Environment` to `NotePlan` 

My thinking there is, future “global” methods could be added which are not Environment type methods. I have proposed to @EduardMe to creature a new module `NotePlan` where he can add `environment` and future methods.

This PR will wait until @EduardMe has made his final decision as to where to place these “global” methods.

m1well:
> Ah, I see what you mean there.   IMHO, I think it would be better to rename `Environment` to `NotePlan` 
> 
> 
> 
> My thinking there is, future “global” methods could be added which are not Environment type methods. I have proposed to @EduardMe to creature a new module `NotePlan` where he can add `environment` and future methods.
> 
> 
> 
> This PR will wait until @EduardMe has made his final decision as to where to place these “global” methods.

But the current code is also wrong. And with this PR we at least avoid the errors

dwertheimer:
Closing this PR because: (a) the understanding of Environment is not correct (the comments are helpful to @EduardMe though) and (b) it rolls back some of the changes that @jgclark made and should persist.

PR #193, opened 2022-01-08
Update helpers/userInput to use new API funcs

jgclark:
Updates getInput, showMessage, showMessageYesNo to use the two new functions.
- Extends the number of parameters, but provides defaults so existing code will work
- Also falls back to the older method if running pre 3.3.2


jgclark:
I'd like your input, too, @m1well, but I can't seem to add you to the list.

mikeerickson:
@jgclark Reviewing this PR on my phone, will look at it later this evening on iPad.   But, from what I am able to view, it looks good!

m1well:
> I'd like your input, too, @m1well, but I can't seem to add you to the list. In particular for getInputTrimmed.

ok you can have with pleasure:
1. why did you change the return of `getInput`/`getInputTrimmed` to `boolen | string` instead of `string`?
2. you changed the parameters of my `getInputTrimmed` but forgot to change the JSDoc
3. one FlowFixMe caught my eye - but maybe there are more of them:
```
  // $FlowFixMe[incompatible-type]
  console.log(`param config: ${dateParams} as ${JSON.stringify(paramConfig)}`)
```
you all know i don't always like to choose the easiest way with all these `FixMe`.  
why don't you think about how you can do it without, e.g.:
```
  console.log(`param config: ${dateParams} as ${paramConfig ? JSON.stringify(paramConfig) : 'N.A.'}`)
```
4. you forgot to change the `multipleInputAnswersAsArray()`
5. on `showMessage()` you store the answer - but this is not needed
6. this `if (typeof CommandBar.textPrompt === 'function')` is a nice idea, but does this work as excpected?

jgclark:
> ok you can have with pleasure:
> 
> 1. why did you change the return of `getInput`/`getInputTrimmed` to `boolen | string` instead of `string`?

Because that's what the new API functions return. Cancel → false.

> 2. you changed the parameters of my `getInputTrimmed` but forgot to change the JSDoc

Fixed

> 3. one FlowFixMe caught my eye ...

Fixed.

> 4. you forgot to change the `multipleInputAnswersAsArray()`

I didn't forget, but left it for you to look over, because there's now an additional title parameter to deal with, and I wasn't sure what you might want to do with that.

> 5. on `showMessage()` you store the answer - but this is not needed

Fixed.

> 6. this `if (typeof CommandBar.textPrompt === 'function')` is a nice idea, but does this work as excpected?

It does in my testing with macOS (3.3.1) and iOS (3.3.2) :-)

jgclark:
Having done a little more work using the new code, I've realised the challenge of this PR, @mikeerickson @dwertheimer @m1well.
It stems from Eduard making the signature of the new function textPrompt() `boolean | string`, the first being the result if the user cancels. This is helpful, but as it stands it will mean flow correctly flags up errors on all our `getInput` calls.

<time passes>

But I have now checked all the available plugin code and see that apart from one call by nmn in the templates, I'm the only one using it.  Which is a bit surprising, but makes this simpler. As I need to touch all my code anyway (for the reasons given above), I will just switch over to using the newer call, rather than attempt to change `getInput()`.  The one remaining call by nmn will get replaced by Templating v2 anyway.

PR #209, opened 2022-01-17
Summaries 0.5.0 for test 2

### LINKED ISSUE #207: Summaries Plugin - saved search highlights inside links
@jgclark I don't know if you would consider this a bug or not, but you may want to deal with it anyway:

When highlighting is turned on and the term is found inside a URL or a link title, the highlighting breaks the links in a couple of different ways:

In these examples, I searched for "tennis":

In this example, the highlighting is not even visible in the Editor, but the link is re-written and now will 404 when clicked:
Screen Shot 2022-01-16 at 8 09 30 AM

In this example, the text is found in the path to the attachment:
Screen Shot 2022-01-16 at 8 09 49 AM

jgclark:
fix #207

### DIFF
--- a/helpers/__tests__/paragraphs.test.js
+++ b/helpers/__tests__/paragraphs.test.js
@@ -1,0 +1,27 @@
+/* global describe, expect, test */
+import * as p from '../paragraph'
+
+describe('summaryHelpers', () => {
+
+  describe('termNotInURL()', () => {
+
+    test('1', () => {
+      const result = p.termInURL('tennis', 'Something about tennis in http://www.tennis.org/')
+      expect(result).toEqual(true)
+    })
+    test('2', () => {
+      const result = p.termInURL('tennis', 'Something about tennis in [tennis](http://www.tennis.org/booster).')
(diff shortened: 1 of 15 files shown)


jgclark:
For @dwertheimer to test

helpers/__tests__/paragraphs.test.js
```diff
@@ -0,0 +1,27 @@
+/* global describe, expect, test */
+import * as p from '../paragraph'
+
+describe('summaryHelpers', () => {
+
+  describe('termNotInURL()', () => {
+
+    test('1', () => {
```
dwertheimer, line 8: love to see the tests! this is no big deal, but generally speaking, it's good for the test description to be descriptive so someone other than you running the tests, when it fails, knows what this (failed) test was meant to test for. descriptions generally start with "should". So in this case, instead of "1", it would say:
`test('should find search term in a URL'...`

dwertheimer (approved):
works great.  looks great. thanks @jgclark. Just making one stylistic suggestion (for the future). This is good to go from my perspective.

PR #220, opened 2022-02-02
Feature/event automations new config

dwertheimer:
- removed getOrMakeConfigurationSection from dwertheimer.EventAutomations
- updated include/exclude to work with CSV strings of keys (were arrays before)
- added onUpdateOrInstall function for new config settings
- requires NotePlan 3.4


EduardMe:
I have tested every available setting in the events automation plugin and this code works now, it's good to merge from my side.

EduardMe (approved):

PR #223, opened 2022-02-05
Refactor/NPCalendar - move shared files so jgclark and I can both use the events-to-calendar code

dwertheimer:
I have tested this code and the refactored code works just like it did before the refactor.
There may be a pre-existing bug though. See Discord.


jgclark:
I'm not sure why this isn't showing some bigger merge conflicts, as I've also changed timeblocks.js before I saw this.

jgclark:
Rats. My changes (on main) have disappeared.

PR #225, opened 2022-02-05
Moved NPevents function to NPCalendar and changed imports

dwertheimer:
refactored to move one function out of NPevents to NPCalendar. Tested code and looks fine.


dwertheimer:
getEventsForDay is the function
only used in jgclark.EventHelpers > eventsToNotes

PR #230, opened 2022-02-10
filer plugin rollup compile error fix

m1well:
there was a missing `}` in the middle of the code...
also there are some other code errors (I will post screenshot in discord) which I can't fix


jgclark:
Thanks. This code is WIP, and the latest commit message should say that.
I'm part-way through updating it for ConfigV2, as well as some other improvements.

jgclark:
Merging anyway ;-)

PR #244, opened 2022-03-14
add url strip helper. Thanks @jgclark !

dwertheimer:


EduardMe:
I have added some code to render links in time blocks (and remove
markdown), so this should get better automatically?

PR #245, opened 2022-03-15
Fix/np templating clean ups

dwertheimer:


mikeerickson:
Already fixed in b8 which is going through some more testing before releasing

Issue #283, opened 2022-05-27
Rationalise some Event-related plugin and commands


jgclark:
We have many excellent commands in plugins. But inevitably they have grown up without planning, as different authors have seen different (but related) needs.  This issue is to discuss doing some rationalisation. An OK outcome is to decide to make no changes, but at least let's discuss them.

| Command | Current Plugin | Future Plugin? |
| ------------ | ------------- | ------------- |
| /time blocks to calendar | Event Helpers | Time Block Helpers (new) |
| /atb | AutoTimeBlocking | Time Block Helpers |
| Create ATB using Presets | AutoTimeBlocking | Time Block Helpers |
| _rest of EventHelpers commands_ | Event Helpers | Event Helpers (i.e. no change) |

In practice, this would mean dropping 'AutoTimeBlocking' unless we can rename it to 'Time Block Helpers'.

Hmm, fewer things to change than I thought.  No matter. What do you think @dwertheimer?

I've not implemented Reminders commands, but I know others are keen to have them. I've left this as an exercise for others, as I don't use reminders myself. But if we do get them, then they could either live in a parallel 'Reminders Helpers' plugin, or extend 'Event Helpers' to 'Events & Reminders'.

And do you have any plans, @nmn @oschrenk @mikeerickson @m1well or others related to Events, Reminders, Time Blocks?

jgclark:
I know you've been busy brewing up other things, but I wondered if you'd had thoughts on this, @dwertheimer?

dwertheimer:
I’m sorry. I forgot to call you about this. Yes I think that makes good
sense. Happy to do the rename and move timeblocks-to-calendar into it. The
problem is that we will need to think about how to not have duplicate
commands in the filesystem since we can’t move or delete directories. Do
you have ideas?

jgclark:
Thanks -- I understand now.
I will need to have a longer think before replying.

jgclark:
So, we have various options.  But I think the simplest is:

Leave plugin IDs as they are but change the display name AutoTimeBlocking →  Time Block Helpers.
- For users using both plugins: almost won't notice change
- For users using EH but not ATB: add an onUpdate dialog saying '/time block... has moved to TBH.' or (with a installed plugins check) '/time block... has moved to TBH. Please install it now if you wish to use /time block... command'
- For us: downside is continuing untidy filenames.

Howzat?

[...]

PR #315, opened 2022-06-27
Add commands to launch URLs in note open tasks

dwertheimer:
Hi @jgclark. I was looking for a function today to open URLs that are on open todo items in my Editor. Felt like "Note Helpers" would be a good place for it, so that's where I wrote it. See what you think.

### DIFF
--- a/jgclark.NoteHelpers/CHANGELOG.md
+++ b/jgclark.NoteHelpers/CHANGELOG.md
@@ -1,6 +1,12 @@
 # What's changed in 📙 Note Helpers plugin?
 For more details see the [plugin's README](https://github.com/NotePlan/plugins/tree/main/jgclark.NoteHelpers/).
 
+## [0.17.0] - 2022-06-26 @dwertheimer
+
+### Added
+- Add new command  `/open todo links in browser`
+
+
 ## [0.16.0] - 2022-06-26
 ### Updated
 - The function `jumpToHeading` (which is used for jumping to headings within the same note) can Now be used via `x-callback-url` by passing the text of the heading in as an argument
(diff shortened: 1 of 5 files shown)


jgclark (commented):
A quick visual code review looks fine.
There's a wider question, though, which I'll come to in the main conversation.

jgclark:
@dwertheimer @nmn cc @mikeerickson 
As this is now the second recent suggested addition to NoteHelpers where my initial response is "it doesn't fit there", perhaps I need to try and explain my thinking.

My sense of "NoteHelpers" was that it would be for things that work at the level of a Note, "EventHelpers" for things to do with Events etc.
Hence it does /new note, /move note, /rename note, /open note, /convert note etc.  Even /jump to heading is still in the context of a whole note, not a line or event etc.

So you can see why my reaction that "/add number of days to dates" and "/open todo links in browser" don't fit here.  The plugin already has 10 commands without these, and many more will start to feel impractical.

But of course I want to see your new commands out there.  So what else could we do instead? Here are some thoughts:

1. I've already discussed with David (in #283) about making a clearer pair of plugins called "Event Helpers" and (new) "Time Block Helpers". 
2. The "Repeat Extensions" works with dates; perhaps rename that to "Dates and Repeats" and put "/generate repeats", "/add number of days to dates" and "/process date offsets" could all go there?
3. Alternatively, "/add number of days to dates" could live in "Date Automations" plugin, which so far has been less about _automations_ and more about _formatting_ of dates.
4. The (currently unmaintained) "Sweep" plugin feels like a subset of a wider set of 'tidy up calendar notes' activities. As does "/open todo links in browser", as that sounds like the underlying motivation for it -- though it will be useful for other reasons too.  I have some other 'tidy up' ideas in mind as well.

PR #319, opened 2022-07-11
Rename ParagraphBridge to CoreNoteFields and migrate common fields to it

nmn:
Create a new type for the common parts of Note and Editor


jgclark (approved):
I'm approving.

PR #322, opened 2022-07-17
Feat/templating add weekly for self-running templates (<THISWEEK> and <NEXTWEEK>)

dwertheimer:
The changes to add writing to a weekly note were a little more extensive than I thought, so I went ahead and made a PR so you could see the changes. Hope you can work with it. Lemme know.


dwertheimer:
I only edited one file: Templating.js
However, the new code calls some new weekly-note functions in @helpers/dateTime.js

PR #329, opened 2022-08-04
Potential fix for ignoring last line + more logs to be sure

dwertheimer:
Pls see discussion: https://discord.com/channels/763107030223290449/1004560263627944016/1004560265330835477

Pasted for posterity:

dwertheimer — Today at 6:23 PM
@jgclark Question for you. I have a "block" like this:
## [Events to Add to Calendar](noteplan://x-callback-url/runPlugin?pluginID=dwertheimer.EventAutomations&command=cevt%20-%20Create%20Events%20From%20Text&arg0=Events%20to%20Add%20to%20Calendar)
Pick up Pierce at 5pm
Take Anique to Wells Fargo/Cava at 3pm
Cancelled: Tennis with Tomer at 6pm

When I call getParagraphBlock(), I get all but the last line.
I looked at the code, and I see why.
  const endOfActiveSection = findEndOfActivePartOfNote(note)

In my case, note.paragraphs[endOfActiveSection].content is equal to the last paragraph in my block. 
So far so good.
But the for loop does this:
    for (let i = startLine + 1; i < endOfActiveSection; i++) {
Note the i < endOfActiveSection
Feels to me like this should be i <= 
That works for me, but I don't know if that will mess up other places you've been happily using this code. So I don't want to push the change.
Also there's another loop which maybe needs adjusting as well.
    // See if there are following indented lines to move as well
    for (let i = startLine + 1; i < endOfActiveSection; i++) {
Can you please have a look and see if you can address it without having some negative consequence for the way you use the code?

### DIFF
--- a/flow-typed/Noteplan.js
+++ b/flow-typed/Noteplan.js
@@ -212,7 +212,7 @@ declare interface TEditor extends CoreNoteFields {
   /**
    * Scrolls to and highlights the given range. If the paragraph is folded, it
    * will be unfolded.
-   * @param {RangeObject} range
+   * @param {Range} range
    */
   highlightByRange(range: Range): void;
   /**
--- a/helpers/NPParagraph.js
+++ b/helpers/NPParagraph.js
@@ -125,21 +125,26 @@ export async function replaceContentUnderHeading(
  * it will work as if the cursor is on the preceding heading line, and then using the same rules as above.
  * (This used to be called 'useExtendedBlockDefinition'.)
  * @author @jgclark
+ * @tests available in jest file
  *
  * @param {Array<TParagraph>} allParas - all selectedParas in the note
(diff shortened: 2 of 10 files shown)


jgclark (commented):
WIP for @dw test

dwertheimer (commented):
@jgclark I don't see any real code changes in this commit. Only a Flow update and a CHANGELOG update. Both are fine/great with me (assuming my PR changes worked for you in your plugins). You just mentioned that there is Filer code using this, but I don't see any changed plugin.json with new settings or anything.

jgclark:
Grrr.  VSC is telling me its committed in the branch, but it's only showing you the files that you changed.
This is why I've tended to avoid PRs.
I'll try again ...

PR #337, opened 2022-08-30
EventHelpers: Fix formatting of *|DATE|* to use the configured locale

akrabat:
In EventHelpers, `*|DATE|*` should be formatted using the locale set in config and if this is empty, then the operating system's locale.

PR #340, opened 2022-08-31
EventHelpers: Add new `includeAllDayEvents` parameter

akrabat:
In both `events()` and `matchingEvents()`, support a new `includeAllDayEvents` boolean parameter to allow disabling the display of all day events by setting to `false`. Default is `true`.

Note that I expect this PR to be merged after #337 as it is branched off that PR's branch.


jgclark:
The only think I've spotted is that the version number (plugin.json) is different from what you put in the CHANGELOG. But I'll fix it.

jgclark (approved):

PR #343, opened 2022-09-07
adding DataStore.settings['_logLevel'] = 'none' to reduce noise

dwertheimer:
@jgclark There was way to much console logging going on in the testing, so I added this line to most of the offending test files so that you could change that setting when you really want to see during the tests, but otherwise, the terminal won't be cluttered with logs when just running tests. I made this as a PR so you can look and make sure it won't mess up anything you are working on. Please let me know.


jgclark (approved):
Ok ...

PR #350, opened 2022-10-24
Fix errors oct 2022

nmn:
Fixes errors so I can then update Flow.


jgclark (approved):
I've not checked every change in my files, as most of the changes appear to be simply 'prettifying'.

np.Templating/src/Templating.js
```diff
@@ -561,9 +562,10 @@ export async function templateWOTD(): Promise<void> {
       },
     }
 
-    const result = await fetch(url, options)
+    const json: mixed = await fetch(url, options).then((res) => res.json())
```
dwertheimer, line 565: @nmn This was one of the few substantial (not just Flow/checks) changes I saw you make. This one does not work. res.json() does not exist in the promise that is returned by NP. The result of NP's fetch is actually a Promise<string>. For example:

```
response :: {"word":"wood ipecac","pronunciation":{"all":"wʊd_'ɪpə,kæk"}}
```

dwertheimer (changes requested):

PR #378, opened 2022-12-20
jgclark.Reviews: Fixing folder of project completion

debone:
The plugin was not respecting the configured folder and not registering the projects completions if you had changed it.


jgclark (approved):
Thanks for spotting and fixing this, @debone

jgclark:
I'm quite a long way into preparing a next release of this plugin, improving some other things. But I can't do that until I'm through with my Christmas work. So don't be surprised if it takes a week to appear.

PR #384, opened 2023-01-04
Frontmatter: Improve colon quotting

aaronpoweruser:
- Quote titles
- Check if the string ends in a colon


dwertheimer:
nice catch @aaronpoweruser. thank you.

PR #398, opened 2023-01-24
Improve quoting behavior

aaronpoweruser:


dwertheimer:
Thx as always @aaronpoweruser! Question for you...do you think we need to whitespace `trim()` the content before those checks? Because it occurs to me that if you sent it a string like `"is not quoted" ` with an accidental space a the end, it would fail the quoted test as written. I'm not looking at the code so maybe the value is trimmed before hitting this function. I'm going to merge, but see what you think. Thanks as always for the improvements.

PR #399, opened 2023-01-25
Fix super minor documentation error

yatil:
In the last line of the birthday example, `{^0d}` is used which would calculate from the date before, which is the posting date for the package. But I assume you don’t want to call on that date, but on the actual birthday date. Hence, removing the `^` character.


jgclark:
Thanks for spotting this, @yatil!

PR #407, opened 2023-02-26
add fractionalDays param + additional tests

dwertheimer:


helpers/dateTime.js
```diff
@@ -384,22 +384,35 @@ export function monthNameAbbrev(m: number): string {
...
  * Note: v2 uses a.moment(b).diff(moment().startOf('day'), 'days') instead
  * @author @jgclark
  * @tests in jest file
- * @param {Date} startDate
- * @param {Date} endDate
+ * @param {string|Date} startDate - if string, must be in ISO format (e.g. '2022-01-01')
+ * @param {string|Date} endDate - if string, must be in ISO format (e.g. '2022-01-01')
+ * @param {boolean} returnFractionalDays (default: false) - if true, return a fractional number of days (e.g. 1.5 for 36 hours)
+ * otherwise, it truncates the decimal part (e.g. 1 for 36 hours)
  * @return {number} - number of days between startDate and endDate (rounded to nearest integer)
```
jgclark, line 398: This now won't always be "rounded to nearest integer"

helpers/dateTime.js
```diff
@@ -384,22 +384,35 @@ export function monthNameAbbrev(m: number): string {
...
   // const res = Math.round((endDate - startDate) / 1000 / 60 / 60 / 24) // i.e. milliseconds -> days
   // return (res === -0) ? 0 : res // handle weird edge case
 
   // v2 method:
   // moment's a.diff(b, 'days') gives the different in days between a and b, with the answer truncated (not rounded)
-  return moment(endDate).startOf('day').diff(moment(startDate).startOf('day'), 'days')
+  if (returnFractionalDays) {
+    return moment(endDate).diff(moment(startDate), 'days', returnFractionalDays)
+  } else {
+    return moment(endDate).startOf('day').diff(moment(startDate).startOf('day'), 'days', returnFractionalDays)
```
jgclark, line 414: This looks right, apart from its then pointless asking for 'returnFractional Days', as it will always deliver an integer answer, won't it?  It's not wrong, but I think it's just a bit misleading.

jgclark (approved):
Couple of minor things; otherwise please release.

helpers/dateTime.js
dwertheimer: I'm not sure I understand what you mean. It does return a fraction if returnFractionalDays is true

helpers/dateTime.js
dwertheimer: ok

helpers/dateTime.js
jgclark: The else clause makes sure the answer will be an integer, surely?
But its not important to the output, so I'll commit.

PR #430, opened 2023-03-24
Change function reference

jgclark:
For @dwertheimer to check, following name change in QuickCapture plugin


dwertheimer:
Merged into main. Thx @jgclark

PR #434, opened 2023-04-03
"NotePlan Online" is the plugin that allows you to publish your notes securely to the web

antony-sklyar:
This is the first version of the plugin, but it's already useful, as it supports password-protected notes sharing, manual update of the shared content and also (at the backend level) display of some custom markdown flavour that became quite popular among NotePlan users (underline, strikethrough, highlight, task and checklist priorities, inline comments).

The code should be simple to understand, but if you need any explanation, I'd be happy to provide it.

Backend code is published in a GitLab public repo and referenced in the README, so you are able to review it too if you want. It is deployed to noteplan.online domain that I bought to use specifically for this project, so that's not some unknown 3rd party.


jgclark:
I will attempt to pull and build later. But for now I've 3 general comments:

1. Well done: from a quick scan of the code, you appear to be using all the scaffolding correctly
2. I know enough about security engineering to know that it's much harder than it looks. I am in no way qualified to test the actual security of this system, particularly as I have no idea about the underlying noteplan.online system. If I really cared about E2E encryption (and TBH I don't) then I'd probably want to know more about it in the documentation.
3. As this is using a (new?) web service with 'noteplan' in the name, I suggest you check with @EduardMe that he is happy with the principle of it. He only wants to support development of new plugins, but this looks to me to break possibly sensitive new ground.

antony-sklyar:
The implementation is quite simple and open-sourced (link is in the README file of the plugin) and it uses encryption algorithms that Laravel PHP framework has out of the box. I did not want to mess around with JavaScript encryption libraries in the plugin code itself, decided that encrypted transfer channel (HTTPS) and encrypted storage (AES encryption before storing in the database, without storing the encryption key on the server) would be quite enough.

As for the backend tied to noteplan.online domain name, I just wanted public URLs to look good, but obviously could not use official domain so I bought another one that was cheap enough and beautiful enough for the purpose.

@EduardMe, if you would feel like you want to host the published notes backend on your own server and/or create a subdomain of noteplan.co for this plugin to use, I would gladly embrace it and cooperate. Just wasn't sure if it's possible, so started with my own domain name and server.

antony-sklyar:
Despite the fact that we've started our first steps on the "Publish" feature improvement with @EduardMe, I thought that this pull request should not hang here forever because of the naming issue, and migrated to a more neutral name `Share Note` for the plugin as well as more neutral domain name `sharednote.space` for the backend.

The only thing that is not yet documented in `README.md` is the new way to obtain API token. I'll document obtaining the token as soon as I have time, and/or implement simplified process that would allow anonymous usage of the plugin without registration on the backend and obtaining the API token.

dwertheimer:
Sounds good  after u do that let’s publish it

antony-sklyar:
Done.

EduardMe:
Sorry, I didn't comment here my feedback, because I thought working on the built-in publishing feature would make this redundant.

What's the general plan with the plugin, should it be available in parallel in the future along with the publish feature? I'm asking because it could add confusion for the user, especially if it has the same features (once the built-in one has password protection, etc). And you can also share notes with other NotePlan users, which is another built-in feature.

- We could add "online" to the name, as in "Share Notes Online" or "Web Note Sharing", because just "Share Notes" suggests you share them privately with other NotePlan users.
- We should add a notice prominently somewhere to warn the user that the content of the note will be uploaded to third-party servers. We do this already in the "NotePlan AI" plugin with a prompt that is shown once (I think), which I liked because there are a lot of users concerned about privacy.

EduardMe:
Just looked through the code and readme. Where is the link to the backend source code? I saw it once, maybe on Discord or so, but I can't find it in the readme?

antony-sklyar:
Observations:

* It seems that I'm bad at naming collisions, switched from bad name with "noteplan" in it to a bad name with "share" in it 🤣
* You're right that I forgot to include the link to the backend code into the plugin README this time. Will fix that.

Regarding the "NP Online / SharedNote" plugin+backend future in general. Based on our latest Discord discussions, it seemed to me that my original idea was slightly different from both "NotePlan Publish" and "NotePlan for web" current vision. This is why I thought that that original idea is still relevant and worth pushing as an independent plugin.

antony-sklyar:
Specifically, with the latest updates to Shared Note Space backend the notes are bound to the user account and the user can be invited to one or more teams.

This, combined with recursive publishing of the notes, will allow to easily provide a "team space" of the shared notes, first with NotePlan as the native editor for those notes (with the ability to publish fresh version of the note via the plugin, and thus also providing version history for the published notes). Then we could provide the capability for (at least limited) editing of the notes online and downloading fresh version of the note from the server down to your NotePlan app.

All of this, combined with secure storage of the notes from the start, and not being bound to Apple ecosystem, lies a seemingly good foundation for collaboration between NotePlan users and non-NotePlan  users, and Apple users with non-Apple users.

These are some of my thoughts, for better mutual understanding.

antony-sklyar:
P.S: But friends, if you would find these ideas as conflicting with NotePlan vision, I can withdraw this PR and keep it just for my private usage, "to scratch my own itch" so to say, and avoid the confusion around the topic of publishing / sharing / web features.

antony-sklyar:
OK, I'm closing this PR then, and will keep you posted if this will grow into something useful or interesting 👍

PR #444, opened 2023-05-06
jgclark.Reviews speed improvements

dwertheimer:
@jgclark Sorry I wasn't complete enough in my last post. Here's some more detail:
Scanning my Project Notes with the latest release takes 3.6s and finds only 10 notes

Here's my analysis:

If you are searching for multiple hashtags, *each hashtag* takes 58ms multiplied times the number of hashtags and folders you search in (e.g. in my case 200ms for each folder) .
For example:
```
findNotesMatchingHashtag(#bevProject, Programming/Projects): 0m0s.58ms
findNotesMatchingHashtag(#project, Programming/Projects): 0m0s.53ms
findNotesMatchingHashtag(#homeProject, Programming/Projects): 0m0s.54ms
findNotesMatchingHashtag(#personalProject, Programming/Projects): 0m0s.52ms
```
That really adds up!

...but I found that nearly *all* of that time comes from this one line:

```js
      projectNotesInFolder = DataStore.projectNotes.slice().filter((n) => getFolderFromFilename(n.filename) === folder)
```

So, filtering the projectNotes is expensive.
I made one small change and that was to filter projectNotes once and then pass that to the function.

```js
    const filteredDataStore = DataStore.projectNotes.filter((f) => filteredFolderListWithoutSubdirs.some((s) => f.filename.startsWith(s)))
    ...
    const projectNotesArr = findNotesMatchingHashtag(tag, folder, false, '', true, filteredDataStore)

```

I can do that filter once in 111ms, rather than paying 53ms for every call

This resulted in a 10x speed increase
After the change, the scan takes 0.33 seconds (instead of the 3.6s before)

With another optimization to the DataStore filtering, I was able to get it to 272ms. I might be able to do even better, but given that it's 1/4 a second, I think I'll stop here. :)

Please try this PR out against your notes and compare it to your current/expected results. It *should* be identical just way faster (it is for me). But pls test to be sure!

### DIFF
--- a/helpers/NPnote.js
+++ b/helpers/NPnote.js
@@ -340,12 +340,25 @@ export function getNoteTitleFromFilename(filename: string, makeLink?: boolean = 
  * @param {boolean?} caseInsensitiveMatch - whether to ignore case when matching (default true)
  * @return {Array<TNote>}
  */
-export function findNotesMatchingHashtag(tag: string, folder: ?string, includeSubfolders: ?boolean = false, tagToExclude: string = '', caseInsensitiveMatch: boolean = true): Array<TNote> {
+export function findNotesMatchingHashtag(
+  tag: string,
+  folder: ?string,
+  includeSubfolders: ?boolean = false,
+  tagToExclude: string = '',
+  caseInsensitiveMatch: boolean = true,
+  notesToSearchIn?: Array<TNote>,
+): Array<TNote> {
+  // logDebug(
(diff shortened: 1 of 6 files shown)


dwertheimer:
BTW I fixed some Jest tests that broke with the previous Reviews update. They all should pass now.

jgclark (commented):
Thanks, @dwertheimer, this is indeed quicker, and appears to behave the same.

PR #456, opened 2023-06-26
new command: /Remove >today tags from completed todos

dwertheimer:
I found that I had several completed tasks that had been tagged with ">today" and were still showing up (forever) in my References section. This command gets rid of the ">today" on completed/cancelled tasks and checklists.


jgclark (changes requested):
Good work, thanks.
To complete it, though, please can you add this command as an option for Tidy Up All (requires new setting + update in tidyMain.js file)?

dwertheimer:
OK @jgclark I added the command to TidyUp. 
Tested it silently and not and works for me

jgclark (approved):

PR #458, opened 2023-07-12
Templating code blocks fix

dwertheimer:
Introduced \`\`\`templatejs tag and allow all other codeblocks in templates


dwertheimer:
@mikeerickson @EduardMe I think we should merge and release this code to fix the bug a user reported.

PR #460, opened 2023-07-16
initial integration for @Leo test and additions

jgclark:


jgclark (commented):
Integrated to v0.18.1

PR #463, opened 2023-07-23
Add Inconsistent file names functionality to Note Helpers plugin

leomeloxp:
Add three new commands to the Note Helpers plugin:

- **list inconsistent note filenames:**  Lists the names of notes whose filenames are inconsistent with their titles
- **rename filename to title:** Renames the current filename to the title of the note
- **rename inconsistent note filenames:** Renames the files of notes whose filenames are inconsistent with their titles

A few things to note:

- Renaming a single note will always prompt the user before performing an action
- Bulk rename can optionally be limited to a single folder (selecting `/` in the dropdown will do the entire list of project notes)
- Bulk rename will always prompt the user before starting the rename process, allowing the user to choose whether they want to do the renaming interactively (ie, being asked per note) or automatically. It will also allow them to cancel the entire operation at this point

These are fairly simple functions but they naturally touch a sensitive type of functionality: renaming the actual notes.

It could be worth adding some mention before doing the bulk renaming that this operation could impact syncing of notes and could lead to different devices falling out of sync for a bit of time if there are too many notes being renamed at once.

There could also be a risk for data loss in this case but I don't really know how robust the sync logic is and/or how it attempts to prevent that. As always, if unsure, make a backup of your notes before attempting to do this.

As a minor improvement to this, I could eventually look into listing all the notes in a HTMLView and offer to allow the renaming directly from there. I did look a bit into how I could make this work but it seems like there's quite a bit of work required to make it happen.

_**NOTE:**_ I also sneaked in a `.tool-versions` file to help with managing node JS versions. I set it to the latest Node 16 version as that's the latest supported version.


leomeloxp:
@jgclark tagging you here as we were discussing it on Discord.

jgclark:
Will be including in NoteHelpers v0.18.1

PR #485, opened 2023-08-26
[NoteProtector] [0.1.0] Initial release

aaronpoweruser:


aaronpoweruser:
Been using internally for a while. Posting for feedback. 
ill circle back and update the readme with cipher details, etc.

aaronpoweruser:
Need to add support for inserting line "%%" to hide the cipher text

Need to add crypto details so can be decrypted using opensssl

Check is a whole note can be encrypted based on a a tag in front matter with the option to ignore headings and maybe tags.

jgclark:
I think I commented on an earlier incarnation of this plugin, under a different name. Anything involving security and particularly cryptography is notoriously difficult to do without intentionally creating vulnerabilities. Without much fuller details in the README about what use cases its designed to support, and what security approach you’re relying on, I wouldn’t want to use this.  That said, at least you appear to be using a 3rd party library that I hope will have had good external scrutiny.

It would also be helpful to add to the README what Eduard has said about what encryption of data at rest and on the wire NotePlan uses with iCloud. (I.e. what is in place, and what this adds.)

Finally, one small comment about emojis I can see in the source code: one appears to be of a closed lock and the other a closed lock with a key. Reading the code quickly appears to suggest that one of them means Decrypt which seems not match either of those symbols. (But of course I could be jumping to the wrong conclusion from the code.)

aaronpoweruser:
Agree this is NOT a secure solution by any means. It is meant to protect the day at rest. If internal details of NotePlan are known this is easily bypassed by various means. 

My use cause is protection of sensitive info. Passwords accounts numbers etc. that I want to store in NotePlan. 

I'll add some details about how this is more of a protection against over the shoulder/physical access attacks than a real secure solution. 

The emojis are just arbitrary tags to indicate protected and to be protected text that are user configurable. 


The end goal is to describe all the cipher details and allow decryption by openslll.

PR #534, opened 2024-02-28
Fix CI builds: Update package-lock.json

aaronpoweruser:
Github actions uses 'npm ci' to build which does not update package.lock which is out of date with package.json causing 'npm ci' to exit leading to CI build failures.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "np.plugins",
-  "version": "3.19.0",
+  "version": "3.20.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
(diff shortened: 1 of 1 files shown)


jgclark:
I don't understand the npm system very well. So in the very rare times I've updated package.json I've never committed package-lock. json because I don't want to break things for other devs.

Question: in general is it necessary and safe to commit package-lock.json as well as package.json?

aaronpoweruser:
Yes, package-lock.json is intended to be committed every time package.json is changed (normally via npm install). See https://docs.npmjs.com/cli/v10/configuring-npm/package-lock-json

PR #536, opened 2024-02-28
Fix test: Add missing setEditorSplitWidth function to WindowTools 

aaronpoweruser:
Not sure what broke here, I was able to verifiy functionality and pull the code from the
published 1.0.0 version, but the method never shows in git.

Searched commit history with "git grep "setEditorSplitWidth" $(git rev-list --all)" and can not
find a committed version


aaronpoweruser:
🟢 Green light

PR #537, opened 2024-03-04
jgclark.repeatExtensions: dueDate calculation adjustment

ameritrash:
adjust new data calculation: when there is no task due date, use the note date if it exists, otherwise default to (current behavior) of task completion date.


jgclark (changes requested):
Thanks for this.
The code and plugin.json changes look fine, but to be complete it also needs documentation updates. In the case of a NotePlan plugin that means the CHANGELOG.md and README.md as well.
Please could you make those (easy) changes as well?

ameritrash:
so, I made the requested changes, but cannot figure out how to submit them to existing PR. or are they already there?

PR #548, opened 2024-03-28
Fix: start fixing type errors

nmn:
Flow hadn't been updated in years (a new version is released every few weeks). The old version of Flow was crashing and type errors have basically been ignored this whole time.

I updated Flow, the type definition files and fixed a few type errors.

There are still over 1000 type errors.

---

Related:

The repo overall can be cleaned up quite a bit. I have learned more about NPM etc in the last couple of years.

However, to make this work, we would need to restructure the repo a bit. I'm not sure if Noteplan can continue to work if the location of plugins changes.


dwertheimer:
Thx again @nmn. Would love to clean up the Flow errors. @nmn what's the best way for us to make sure this all works? Obviously the ones that are just flow annotations are not an issue, but the ones that are actual code changes could have effects. The tests all still pass, which is good, but much of the plugin code does not have tests because of the Noteplan sandboxing. So what would you suggest? It would be fantastic to get this down to zero errors. I'm very willing to help/try (with my limited knowledge of Flow).

@jgclark I see that the CI build is failing in this branch. Can you please check it out?
```
Build of plugin: "/Users/runner/work/plugins/plugins/jgclark.NoteHelpers" failed Error: 'toLocaleDateString' is not exported by helpers/NPdateTime.js, imported by jgclark.NoteHelpers/src/indexFolders.jsBuild of plugin: "/Users/runner/work/plugins/plugins/jgclark.NoteHelpers" failed Error: 'toLocaleDateString' is not exported by helpers/NPdateTime.js, imported by jgclark.NoteHelpers/src/indexFolders.js
```

jgclark:
@dwertheimer I've hopefully fixed CI error.

@nmn I'm going through the many changes. So far: 
- Thanks for prettier update.
- Can you also update package.json to get us up to whatever flow version you suggest?

nmn:
> Can you also update package.json to get us up to whatever flow version you suggest?

@jgclark I increased the Flow version by the minimum amount that would get it to stop crashing for now. Ideally we should track the latest version at all times, but since it'll reveal and more and more errors we should fix the type errors before trying to update further.

The `package.json` file is already updated to version `0.210.0`.

> Would love to clean up the Flow errors. @nmn what's the best way for us to make sure this all works? 

@dwertheimer After this change Flow is no longer crashing so please try and fix as many errors as possible. Remember that you can use the `any` type wherever you feel like to tell Flow to shut up!

But also, I want to stress that if you're not getting any value from the types and it's only getting in your way, I would suggest changing the `// @flow` at the top of your file to `// @noflow` instead. This will turn turn off all type-checking without forcing you to remove the type annotations from the code.

nmn:
feel free to merge it?

PR #557, opened 2024-04-12
Readwise sync log

aaronpoweruser:


aaronpoweruser.ReadwiseUnofficial/src/NPReadwise.js
```diff
@@ -34,6 +34,9 @@ async function handleReadwiseSync(response: any): Promise<void> {
   await response.map(parseBookAndWriteToNote)
   log(pluginJson, `Downloaded ${downloadHiglightCount} highlights from Readwise. Updated ${updatedSourceCount} notes.`)
   await showMessage(`Downloaded ${downloadHiglightCount} highlights from Readwise. Updated ${updatedSourceCount} notes.`)
+  if (DataStore.settings.writeSyncLog === true) {
+    await writeReadwiseSyncLog()
```
aaronpoweruser, line 38: My JS is failing me. The update count reports 0 because the parsing map has not returned yet. it is tagged with await, so not sure what the issue is. 

The value is always correct for the second caller so if "showMessage" and "writeReadwisezSyncLog" is inverted the first one reports 0 updates and the second the correct value.

aaronpoweruser:
Ready for merge.

jgclark (approved):
Not tested, but visually reviewed.

PR #567, opened 2024-04-24
Summaries 0.22.1

aaronpoweruser:


jgclark:
I'm happy with the tidy up and documentation changes.
But I'm not fully understanding the change to `maxTermLen`. If I understand the code right, then this will mean having a hugely long (>50 character) hashtags or mentions will stop padding, and make all of the progress update essentially left-aligned without padding. This will surely look ugly?
Have you considered the alternative which is to internally truncate the very long hashtag/mentions?  Eg. "#thisIsSoLongThat...APR" instead of "#thisIsSoLongThatItsCausingWritingOfAPR".

(Longer term I want to be able to generate and insert proper graphical sparkline graphs here instead of the 'ASCII Art', but the necessary support isn't really available in NP yet.)

aaronpoweruser:
Your right truncating will look better. Ill make a new ps.

jgclark:
Thanks for checking with me on this.
Feel free to release the version with shortening without checking in with me first.

aaronpoweruser:
Checklist padding fix 

There is some weirdness with rendered text link links but this is good enough for now.

dwertheimer:
Ah yes, nice catch. we have seen this issue in the main app also. I would suggest filtering out links and replacing with the text in the brackets before truncating. @jgclark do we have a helper for this? We certainly have code which could be helpful when we convert markdown links to HTML links.

dwertheimer:
@aaronpoweruser ^^

aaronpoweruser:
Done, while I was there added template render shoutout @mikeerickson for good code. 

Code looks right however I am running into a bug I have seen before were "await" is not honored and code execution continues before the promise is completed. I thought it was related to date returned by Np but it seems its on the 
Node side as this code is pure JS and does not use the Np api.


```
2024-04-24 18:03:42 | INFO  | gatherOccurrences :: Gathered CompletedChecklistItems data in 19ms
2024-04-24 18:03:42 ❗️ERROR❗️ generateProgressUpdate :: this should be after the await
2024-04-24 18:03:42 | INFO  | makeProgressUpdate :: Appended progress update for this week to current note
2024-04-24 18:03:42 ❗️ERROR❗️ generateProgressUpdate :: rendered term:  💭 Plan day [review 2024-W17](noteplan://x-callback-url/openNote?noteTitle=2024-W17)
```

jgclark.Summaries/src/summaryHelpers.js
```diff
@@ -720,10 +724,19 @@ export async function generateProgressUpdate(occObjs: Array<TMOccurrences>, peri
...
     // Include sparklines only if this period is a month or less
     const showSparklines = (requestToShowSparklines && daysBetween <= 31)
-    if (showSparklines) {
-      // Truncate lines to avoid line wrapping
-      occObjs.forEach((m) => m.term = truncateString(m.term, 60, 'middle'))
-    }
+
+    await occObjs.forEach(async (m) => {
+      if (m.term.includes('<%') && m.term.includes('%>')) {
+        m.term = await NPTemplating.render(m.term)
```
aaronpoweruser, line 730: this is returning late

jgclark: I'm not sure what you're trying to achieve here when you say simply "render templates".
Nor do I know the Templating code, and as it's not well documented, it would take me a very long time to figure out. So I can't help. Do you want to back out this part of the PR for now?

dwertheimer: Looks pretty cool to me. You could include template code in the title. Sounds like it could be useful. @aaronpoweruser What's your use case? And you tested it I assume.

jgclark:
> Code looks right however I am running into a bug I have seen before were "await" is not honored and code execution continues before the promise is completed.

@dwertheimer I wonder if this is related to the issue I'm seeing where commands in templates are returning `[object Promise]` for me?

jgclark.Summaries/src/summaryHelpers.js
dwertheimer: @aaronpoweruser Can you pls provide some test examples that show how your new code is used, especially the new template parts? I don't see any mention of the template strings in the README. I may have missed it.

aaronpoweruser: Sure, I can update the documentation, this was a for me feature. As I dynamically generate a link to my weekly note.

PR #621, opened 2025-02-05
Dashboard overwrite empty sections

dwertheimer:
@jgclark Pls have a look. I found (as noted in the Dashboard note) that when you click on the last item in a section the section does not go away. I think this is because the section is not getting overwritten in the refresh because items is blank and therefore a section is not produced and therefore the old section is not overwritten. I went through and changed this in all the dataGens I could find. I took out the ( if items.length > 0 then push section) parenthetical and add the section in all cases so it will overwrite the old one with a new section with no sectionItems on refresh. 

Please test in this branch. 
- click done on all items in a section
- the row should fade away
- the section should then disappear when it is replaced with a new section that has empty sectionItems array (so it's there but doesn't display)


jgclark:
Merging branch after visual inspection

PR #624, opened 2025-02-17
Moved Content to HTML functionallity to Shared API and implement IMG files

cwhittl:
Moved Content to HTML functionallity to Shared API and implement IMG files to base64

### DIFF
--- a/helpers/HTMLView.js
+++ b/helpers/HTMLView.js
@@ -115,7 +115,6 @@ export async function getNoteContentAsHTML(content: string, note: NPNote): ?stri
     // In particular remove trigger line
     if (hasFrontmatter) {
       let titleAsMD = ''
-      lines = lines.filter(l => l !== 'triggers: onEditorWillSave => np.Preview.updatePreview')
       // look for 2nd '---' and double it, because of showdown bug
       for (let i = 1; i < lines.length; i++) {
         if (lines[i].match(/^title:\s/)) {
--- a/helpers/HTMLView.js
+++ b/helpers/HTMLView.js
@@ -97,8 +97,9 @@ export function getCallbackCodeString(jsFunctionName: string, commandName: strin
 
 
 /**
- * Convert a note's content to HTML
+ * Convert a note's content to HTML and include any images as base64
  * @author @cwhittl
+ * @param {string} content
  * @param {NPNote} Note
  * @returns {string} HTML
  */
--- a/helpers/HTMLView.js
+++ b/helpers/HTMLView.js
@@ -98,12 +98,11 @@ export function getCallbackCodeString(jsFunctionName: string, commandName: strin
 
 /**
  * Convert a note's content to HTML and include any images as base64
- * @author @cwhittl
  * @param {string} content
- * @param {NPNote} Note
+ * @param {TNote} Note
  * @returns {string} HTML
  */
-export async function getNoteContentAsHTML(content: string, note: NPNote): ?string {
+export async function getNoteContentAsHTML(content: string, note: TNote): ?string {
   try {
     let lines = content?.split('\n') ?? []
 
--- a/helpers/HTMLView.js
+++ b/helpers/HTMLView.js
@@ -4,12 +4,16 @@
 // by @jgclark, @dwertheimer
 // Last updated 2024-11-09 by @jgclark
 // ---------------------------------------------------------
-
+import showdown from 'showdown' // for Markdown -> HTML from https://github.com/showdownjs/showdown
+import {
+  hasFrontMatter
+} from '@helpers/NPFrontMatter'
+import { getFolderFromFilename } from '@helpers/folders'
 import { clo, logDebug, logError, logInfo, logWarn, JSP, timer } from '@helpers/dev'
 import { getStoredWindowRect, isHTMLWindowOpen, storeWindowRect } from '@helpers/NPWindows'
 import { generateCSSFromTheme, RGBColourConvert } from '@helpers/NPThemeToCSS'
 import { isTermInNotelinkOrURI } from '@helpers/paragraph'
-import { RE_EVENT_LINK, RE_SYNC_MARKER } from '@helpers/regex'
+import { RE_EVENT_LINK, RE_SYNC_MARKER, formRegExForUsersOpenTasks } from '@helpers/regex'
 import { getTimeBlockString, isTimeBlockLine } from '@helpers/timeblocks'
 
 // ---------------------------------------------------------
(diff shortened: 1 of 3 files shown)


helpers/HTMLView.js
```diff
@@ -91,6 +95,148 @@ export function getCallbackCodeString(jsFunctionName: string, commandName: strin
 `
 }
 
+
+/**
+ * Convert a note's content to HTML and include any images as base64
+ * @author @cwhittl
+ * @param {string} content
+ * @param {NPNote} Note
```
jgclark, line 103: type is TNote

helpers/HTMLView.js
```diff
@@ -91,6 +95,148 @@ export function getCallbackCodeString(jsFunctionName: string, commandName: strin
 `
 }
 
+
+/**
+ * Convert a note's content to HTML and include any images as base64
+ * @author @cwhittl
+ * @param {string} content
```
jgclark, line 102: Why pass content and note? Is this different content from what the note has?

helpers/HTMLView.js
```diff
@@ -91,6 +95,148 @@ export function getCallbackCodeString(jsFunctionName: string, commandName: strin
 `
 }
 
+
+/**
+ * Convert a note's content to HTML and include any images as base64
+ * @author @cwhittl
```
jgclark, line 101: Really?

np.Preview/src/previewMain.js
```diff
@@ -82,41 +74,11 @@ Button a { text-decoration: none; font-size: 0.9rem; }
  * @author @jgclark
  * @param {string?} mermaidTheme name (optional)
  */
-export function previewNote(mermaidTheme?: string): void {
+export async function previewNote(mermaidTheme?: string): void {
```
jgclark, line 77: You've removed a default for mermaidTheme. Please add`='green'`.

jgclark (changes requested):
Overall, makes good sense to shift the code around.
Some small points to fix, please.

helpers/HTMLView.js
cwhittl: Updated

helpers/HTMLView.js
cwhittl: The reason is for using it with either a whole note or selected text.
I guess I could just call the editor from the function instead of passing anything?
image

helpers/HTMLView.js
cwhittl: You mean because i wrote a few lines it's not my code?  I 100% don't know what I was thinking, removed.

np.Preview/src/previewMain.js
cwhittl: done

dwertheimer:
@cwhittl @jgclark I pulled and tried to test the PR but I wasn't 100% sure what to test.
I ran it on a test document I had and a bunch of things don't look quite right to me, but I'm not 100% sure what you guys would expect to be covered. I think some things (e.g. checklists etc.) may have been added since jgclark created the Preview plugin. Also some things in NotePlan are not quite standard MD so they don't look quite right.

Also I wonder if there's a way we can use tailwind and the icons in CSS or embedded to improve the HTML a bit?

Screen Cap 2025-02-18 at 19 19 04@2x

[Theme_Builder.md](https://github.com/user-attachments/files/18858049/Theme_Builder.md)

dwertheimer:
@cwhittl @jgclark I wonder if you can bake @EduardMe's custom noteplanstate font into the html itself? ChatGPT says we can.

https://chatgpt.com/share/67b55c18-1244-8012-a635-2d8f9e3c123d

Here are the font files:

[noteplanstate_formats.zip](https://github.com/user-attachments/files/18858785/noteplanstate_formats.zip)

Here are the ascii chars that generate each glyph as far as I can tell:
```javascript
const states = [
  { type: 'open', icon: '*', class: 'noteplanstate' },
  { type: 'done', icon: 'a', class: 'noteplanstate' },
  { type: 'scheduled', icon: 'b', class: 'noteplanstate' },
  { type: 'cancelled', icon: 'c', class: 'noteplanstate' },
  { type: 'list', icon: '-', class: 'noteplanstate' },
  { type: 'checklist', icon: '+', class: 'noteplanstate' },
  { type: 'checklistDone', icon: 'd', class: 'noteplanstate' },
  { type: 'checklistCancelled', icon: 'e', class: 'noteplanstate' },
  { type: 'checklistScheduled', icon: 'f', class: 'noteplanstate' },
]
```

Also in case anybody is interested for posterity, there's some Eduard -> Jonathan discussion [here](https://discord.com/channels/763107030223290449/1034168179464880148/1113502020612136981) about FontAwesome glyphs.

dwertheimer:
Now that I look at [this document](https://noteplan.co/n/56CB53B7-8885-4815-AD79-B90AE344636F) I realize... @eduardme has gone to great lengths to make the HTML version of a published note look as much like it does in the editor as possible. Maybe you guys want to look at whether you can adapt his JS code to generate the HTML preview rather than using the more general purpose showdown? Just a thought...

Of course, @cwhittl still needs @EduardMe to take a look at getting the HTML to clipboard.

cwhittl:
@jgclark @dwertheimer @EduardMe 

I'm really fine either way, my goal has been to be able to copy an HTML version of the a note including images to other applications.   With this change to preview I can do that but completely ok switching gears and looking in another direction if you all want to point me there.

dwertheimer:
I will leave that to @jgclark. But are you able to copy to the Clipboard?

cwhittl:
Right now to make it work I just do a preview and then copy and paste from that window.  On my other plugin that I am making the clipboard doesn't work other than just Clipboard.string from what I have tested.

EduardMe:
Copying html into the clipboard through the API is fixed for the next update.

jgclark:
IMHO we now have a couple of different design questions, and at least one bug, here.

- [ ] 1. Bug: hopefully EM has fixed it already.
- [ ] 2. Decide which plugin should the new functionality live in. Ideally its baked into the app itself, but until then, I think NoteHelpers is the most obvious location, as its a note-level command. Alternatively we change the name of Preview plugin to something that would naturally cover this as well.  [@cwhittl FYI it's easy to change the display name of a plugin, but not the underlying pluginID.]
- [ ] 3. How far to improve the HTML output? When I wrote Preview some while ago, I said I wasn't going to put lots of effort into perfecting the output, as @EduardMe was already working on a web version which should be able in time to give us an 'official' (and therefore maintained) HTML output function to use.

jgclark:
On 2: I'm happy to try this: perhaps "HTML Output" with commands "Save as HTML" (the one one) and "Preview as HTML" (the existing one). Thoughts?  

On 3: I hope @EduardMe can comment on whether getting an API call like `returnHTMLForNote(note: TNote): string` is likely.

EduardMe:
An API call like `returnHTMLForNote(note: TNote): string` is not planned so far.

cwhittl:
So this PR was to make images work with Preview and to move that functionality to a shared place to be used by other plugins.  Are we ok with that?

The bug that @EduardMe has put a fix for doesn't keep this PR from going it will be a problem for my extension (that's not on this PR) and will test it during that process.

jgclark:
I know I was going beyond the scope of the PR with the discussion above, but wanted to capture it while I was thinking about it.
So I will now merge it.

PR #625, opened 2025-02-20
Paperwork - forgot to bump the version, add to the change log, and handle drawing

cwhittl:
Paperwork - forgot to bump the version and add to the change log
Functionality -If an image is a drawing we need to look at the png and give it a white background so it can be read on darkmode.


dwertheimer:
I merged into main. @jgclark you may want to do a release of Preview

PR #631, opened 2025-03-18
add trigger as well as make silent run more silent'er

cwhittl:
Introduces a trigger capability that allows generating repeats for notes from the past four days when working within a calendar note. This feature is not recommended for regular notes, refer to the repeat extension for that use case. 
The trigger line is:
`triggers: onEditorWillSave => np.Tidy.generateRepeatsFromRecentNotes.`

### DIFF
--- a/np.Tidy/CHANGELOG.md
+++ b/np.Tidy/CHANGELOG.md
@@ -1,6 +1,9 @@
 # 🧹 Tidy Up Changelog
 See Plugin [README](https://github.com/NotePlan/plugins/blob/main/np.Tidy/README.md) for full details on the available commands and use from callbacks and templates.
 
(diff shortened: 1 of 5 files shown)


jgclark (changes requested):
I still think adding this trigger isn't the best way to solve this particular problem -- instead use a template to call existing command on daily or weekly notes -- but I'm happy to include it.  It does have the advantage of running more frequently for most people, though it has the down-side of lots of repetitive work that may slow the app down.

Changes requested:
- document in the README
- why hard-code to 4 days?  There's a "How many days to count as recent?" setting which I think you should re-use here. Though I realise its description will need changing a little as this is not a "Remove ... recent ..." command.

dwertheimer:
@cwhittl Do you want to keep the conversation going here or should we close the PR?

PR #655, opened 2025-07-21
NotesFactory and Tasks function

mlevison:
This is the start of the refactoring I proposed in discord.

I created a NotesFactory - since some of the examples are borrowed from helpers/__tests__/paragraph.test.js - it would easy to refactor.

I added tasks.js to helpers and I intend to migrate the other task related functions there as I find them. I.E. from sorting, getAllTasksByType()


__mocks__/factories/noteFactory.js
dwertheimer, line 1: @mlevison This is a great start. Thanks! A starting observation:
- For generalized use, the term "note" is misleading here, because these objects you created are not notes, they are objects with paragraphs arrays. In your tests, you are just accessing that one attribute, which makes sense. However the term "note" is an important distinction because if I wanted to construct a test, I would expect a note to have a title, type, filename, frontmatterAttributes, etc. This is why I was suggesting you use JEST helpers to give you a quick/easy way to create basic note attributes in JSON form.
- Also, defining them all in a single file may get unwieldy if we add lots and lots of them over time. you may want to search the code for the string "const factory = async" to see how factory files are loaded dynamically in Templating by Mike. this way, each test file is actually its own file and does not have to be concatenated in a never-ending single file. That said, I don't like that the Templating factory loader code is import code is repeated. should definitely be a shared function. 
There are a always a lot of ways to deal with these two things. Look forward to hearing your thoughts.

mlevison:
Happy to make improvements.
- I will tweak the note stuff. Where would I find an example of: "This is why I was suggesting you use JEST helpers to give you a quick/easy way to create basic note attributes in JSON form." - since I'm new(ish) to JavaScript I need an example.
- Single file - unweildy true. However, my approach is to refactor when it gets big, that is when I will know what the parts are.

dwertheimer:
@mlevison I'm totally with you on refactoring once a file gets large, but please do have a look at the factories loading in the np.Templating plugin per my previous comment. I think it would be better architecturally to have each note JSON as its own file and then dynamically load them by name when needed rather than having all the notes as a const in one file.

dwertheimer:
@mlevison I just realized that frontmatter attributes was not yet included in the output from JestHelpers so I am pushing that to main. please pull the latest from main and update your branch with this new version.

jgclark (commented):
I've added one comment, about 'tricky' content, but otherwise I'm OK with it so far.

PR #686, opened 2025-09-15
Update for todoist API change

dbludeau:
Change to how results and being handled and fix for query string for Todoist Teamspaces.


dbludeau.TodoistNoteplanSync/CHANGELOG.md
dwertheimer, line 1: @dbludeau I may just be missing it but I don't see a `plugin.json` file in the PR that would include the version 0.4.0 matching what you wrote in the changelog:
```
 ## [0.4.0] - 2025-09-15 (dbludeau)
- Updated plugin to use new Todoist V1 API.
```

Without the plugin.json with a new version, the update will not release -- it gets rejected as no-change.

You should be able to add the changed file pushing to your branch and it will update this PR.

dwertheimer:
@dbludeau Released and merged

PR #687, opened 2025-09-16
Noteplan flow context comment updates

dwertheimer:
@jgclark I often search for commands in NotePlan.js but when I find them don't know if the command is under NotePlan or DataStore or what high-level object. I also find it sometimes hard to read which comment applies to which command. So I asked Cursor to add the full object.command to the JSDOC so that it is immediately obvious when you find one. It's purely optics change, but wanted you to be ok with it since you own this doc. Pls feel free to merge or reject.


jgclark:
Thanks @dwertheimer, for thinking of this useful improvement. Merging.

PR #696, opened 2025-11-04
Preview styling improvements (tables, HRs etc)

dwertheimer:


jgclark (approved):
Thanks, David. My only comment is to suggest you move the new hexToRgb() function to live in helpers/colors.js, so that can be re-used more easily.
(Come to think of it, I think you've used some sort of chroma library in the past. Does that not have this sort of function already?)

dwertheimer:
@jgclark Nevermind this PR. I am going to start a new one

PR #699, opened 2025-11-07
np.Preview: improve wiki links and preview styling

dwertheimer:
## Summary
- convert [[wikilinks]] in preview output into real NotePlan callback URLs (aliases, headings, block IDs supported)
- move preview styling into dedicated file and refresh typography, spacing, list layout, and link presentation
- bump plugin version to 1.0.0 and document the changes

### DIFF
--- a/helpers/HTMLView.js
+++ b/helpers/HTMLView.js
@@ -169,6 +169,8 @@ export async function getNoteContentAsHTML(content: string, note: TNote): Promis
       }
     }
 
+    lines = addParagraphBreakHints(lines)
+
     // Make this proper Markdown -> HTML via showdown library
     // Set some options to turn on various more advanced HTML conversions (see actual code at https://github.com/showdownjs/showdown/blob/master/src/options.js#L109):
     const converterOptions = {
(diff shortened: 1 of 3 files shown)


dwertheimer:
## [1.0.0] - 2025-11-07 @dwertheimer

Arbitrary elevation to 1.0.0 

### General HTML Helper Change
- convert `[[wikilinks]]` in HTML output to be real NotePlan callback URLs rather than just underlined text that does nothing
- fix bug in showdown relative to NP to ensure horizontal rules have a blank line before them so they render properly

### Preview-specific Changes
- move preview-specific spacing and typography to `previewStyles` file, with:
  - table formatting
  - body padding
  - h1 line-height
  - list indentation/line-height tweaks
  - internal note links (no underline, bold on hover)  
  - adjust horizontal rule appearance

dwertheimer:
@jgclark one thing that bugs me but is more fundamental:

I have a wiki link then a return and some other text. Of course, in NP these are separate "Paragraphs" but in the HTML creation, they are combined into a single p tag for reasons I can't really follow. This makes creating a styling rule difficult, because I can't see why they are grouped together. I guess maybe showdown defaults to commonmark standards -- everything is together until you hit two blank lines? I don't know. but it feels to me that this is not the NotePlan assumption so things don't look right.

Screen Cap 2025-11-07 at 08 40 22@2x

If you can think of an easy way to make this look more like NotePlan, that would be great.

A similar thing is happening in this case:

Screen Cap 2025-11-07 at 09 01 20@2x

I have a list, and then a new line that is not part of the list, but the converter. adds a <br> and includes this item inside the list so it gets indented.

If you have any great ideas on what to do here, let me know.

PR #701, opened 2025-11-12
Add section reordering feature to Dashboard plugin

jgclark:
The following done under my direction by Cursor:
- Introduced a new hidden setting for custom section display order in `dashboardSettings.js`.
- Created `SectionOrderDialog` component for visual drag-and-drop reordering of sections.
- Updated `sortSections` function to accommodate custom order logic.
- Integrated section reordering into the `SettingsDialog` for user access.
- Added CSS styles for the new dialog and updated existing styles for consistency.
- Updated relevant types in `types.js` to include custom section order.
- Adjusted the default section display order in `constants.js` for clarity.

### DIFF
--- a/jgclark.Dashboard/src/clickHandlers.js
+++ b/jgclark.Dashboard/src/clickHandlers.js
@@ -509,6 +509,14 @@ export async function doDashboardSettingsChanged(data: MessageDataObject, settin
           return handlerResult(res)
         } else {
           clo(diff, `doDashboardSettingsChanged: Setting perspective.isModified because of changes to settings: ${Object.keys(diff).length} keys: ${Object.keys(diff).join(', ')}`)
+          Object.keys(diff).forEach((d) => {
+            logDebug(
+              `doDashboardSettingsChanged`,
+              `activePerspDefDashboardSettingsWithDefaults['${d}']=${d ? activePerspDefDashboardSettingsWithDefaults[d] : ''} vs. sent to save: cleanedSettings['${d}']=${
+                d ? cleanedSettings[d] : ''
+              }`,
+            )
+          })
         }
         // ignore dashboard changes in the perspective definition until it is saved explicitly
         // but we need to set the isModified flag on the perspective
--- a/jgclark.Dashboard/src/react/components/Dashboard.jsx
+++ b/jgclark.Dashboard/src/react/components/Dashboard.jsx
@@ -8,17 +8,17 @@
 //--------------------------------------------------------------------------
 // Imports
 //--------------------------------------------------------------------------
-import React, { useEffect, useRef, useMemo } from 'react'
+import React, { useEffect, useRef, useMemo, useCallback } from 'react'
 import useRefreshTimer from '../customHooks/useRefreshTimer.jsx'
 import useWatchForResizes from '../customHooks/useWatchForResizes.jsx'
 import { dontDedupeSectionCodes, defaultSectionDisplayOrder, sectionPriority } from '../../constants.js'
 import { copyUpdatedSectionItemData } from '../../dataGeneration.js'
 import { findSectionItems } from '../../dashboardHelpers.js'
 import { dashboardSettingDefs, dashboardFilterDefs } from '../../dashboardSettings.js'
-import type { TSection } from '../../types.js'
+import type { TSection, TActionButton } from '../../types.js'
 import { useAppContext } from './AppContext.jsx'
 import Dialog from './Dialog.jsx'
-import { getSectionsWithoutDuplicateLines, countTotalSectionItems, countTotalVisibleSectionItems, sortSections, showSectionSettingItems } from './Section/sectionHelpers.js'
+import { getSectionsWithoutDuplicateLines, countTotalVisibleSectionItems, sortSections, showSectionSettingItems } from './Section/sectionHelpers.js'
 import { calculateMaxPriorityAcrossAllSections } from './Section/useSectionSortAndFilter.jsx'
 import Header from './Header'
 import IdleTimer from './IdleTimer.jsx'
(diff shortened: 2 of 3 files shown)


dwertheimer:
@jgclark This looks really cool and works! 

A few issues/observations:
a) dragging things to the top doesn't give the right visual feedback. When I drag something, I can only see the blue line underneath another item, but if I'm trying to move it to the top, I don't get the blue line feedback above the first item, and it's not clear to me where the drop zone is. sometimes the drop works (even though i saw no blue line) and sometimes it doesn't.
b) I click "Save" on re-order sections, and I would have expected the Dashboard Settings dialog underneath to also require a save. But it is still greyed out. So I have to click "Cancel" to get out. That seems strange to me as a user. It did actually save underneath, but I was surprised as a user that it did after I clicked "Cancel". I would suggest the re-ordering dialog should say "OK", not "Save" and the Dashboard Dialog should be the thing that saves.
c) Minor styling notes:
- In the re-ordering popup, the "grabber" and the section title have three different cursors, so moving from one to the other changes cursor from finger to hand to crossed arrows. That seems confusing because they all seem to do the same thing. I would think that these would all be one div and it would have one cursor (arrows), but in reality it should be just the up/down arrow (`ns-resize`) since there is no left-right adjustment.

On the main Settings Dialog:

Screen Cap 2025-11-13 at 09 26 03@2x

- space in the button after the icon
- there is too much space above the button
- the description should be closer to the button to match the others. 
- I think it would be more consistent to put this button to the right behind some text like the Max Number of Items (aka the jgclark "compact display")

Maybe it says:
**Section Ordering**: [ Button ]

Hope this is helpful. Again, very cool. Thanks for embarking upon this!

dwertheimer:
@jgclark 

Revised comments:

I love this inline version. So cool!

Dragging feedback:
I liked the blue line concept better even though it didn't work 100% right -- I had written on the previous version "dragging things to the top doesn't give the right visual feedback. When I drag something, I can only see the blue line underneath another item, but if I'm trying to move it to the top, I don't get the blue line feedback above the first item, and it's not clear to me where the drop zone is. sometimes the drop works (even though i saw no blue line) and sometimes it doesn't."

This new version where it highlights an item in the list is confusing to me as a user because I don't know what an item in the list being highlighted means relative to the thing I am dragging. It should either show a line where it's gonna drop it or it should visually re-order the list with the items reordering in real-time as I drag. I think it could do this because there are not many items in the list to re-draw.

Saving:
I had said on the last one: "Save" on re-order sections, and I would have expected the Dashboard Settings dialog underneath to also require a save. But it is still greyed out. So I have to click "Cancel" to get out. That seems strange to me as a user. It did actually save underneath, but I was surprised as a user that it did after I clicked "Cancel". I would suggest the re-ordering dialog should say "OK", not "Save" and the Dashboard Dialog should be the thing that saves.

The latest version now does require two saves but I would have thought that if i didn't save the 2nd one, my changes would not have persisted. But they are immediately sent/written. I would think that "Apply" would be more appropriate, and apply would collapse the inline component and then save on the Settings Dialog. **Or better yet: just get rid of the "Apply" or "Save" on the component altogether and save the result in the Settings Dialog like all other items when they are changed. So when an item is dragged and dropped, it delivers the change to the Settings Dialog every time as a dirty change that needs saving. This feels more consistent.**

Minor styling notes:
- In the re-ordering section, the "grabber" and the section title have three different cursors, so moving from one to the other changes cursor from finger to hand to crossed arrows. That seems confusing because they all seem to do the same thing. I would think that these would all be one div and it would have one cursor (arrows), but in reality it should be just the up/down arrow (`ns-resize`) since there is no left-right adjustment.

jgclark:
OK, try now. (I couldn't remove one of the cursors yet.  It will take longer to figure out quite where it's coming from.)

helpers/react/DynamicDialog/DynamicDialog.jsx
```diff
@@ -44,6 +44,8 @@ export type TSettingItemType =
   | 'button'
   | 'button-group'
   | 'calendarpicker'
+  | 'hidden'
+  | 'sectionOrderPanel'
```
dwertheimer: This is in @helpers/.../DynamicDialog which makes sense, but it's referencing a component that exists only in Dashboard, which is not great.
Feels to me that the list ordering component should be generic and part of DynamicDialog, not specific to Dashboard.

jgclark.Dashboard/PLAN_Section_Ordering.md
```diff
@@ -0,0 +1,267 @@
+# Plan: Visual Section Ordering for Dashboard Plugin
```
dwertheimer: You should probably delete this file before the merge

jgclark.Dashboard/src/react/css/SectionOrderPanel.css
dwertheimer: ofc if you move the code to DynamicDialog, you would want to move much of this CSS also.

jgclark:
Thanks for tracking that down.

helpers/react/DynamicDialog/DynamicDialog.jsx
jgclark: Agreed. 
(One of the challenges of Cursor doing a lot of it is that I was less aware of where the files were that it was touching.)

dwertheimer:
@jgclark Yes, it seems really strange.

If you look at the HTML on the page, it's very odd indeed.

Screen Cap 2025-11-13 at 20 50 14@2x

The settings dialog modal is inside the .headerActionIconButtons
This must be a mistake, but it's been working because the SettingsDialog is absolutely positioned and so it kind of doesn't matter where it emanates from. But it make no sense architecturally.

--

Why does it apply to this <i> element?

Chat says: Summarizing:

.headerActionIconButtons > * matches all direct children (like buttons) of the .headerActionIconButtons element.

The <i> tags inside those buttons are nested descendants, not direct children.

Styles applied to the direct children can cascade down and affect nested elements visually.

If you see the <i> styled, it’s because it inherits or receives styles from its direct parent element (which is a direct child of .headerActionIconButtons).

This explains why the <i> looks styled despite not being a direct child itself.

If you want to style only the direct children but exclude nested elements like <i>, you would need to be specific in your CSS and avoid applying inheritable styles that cascade undesirably.

---

David's translation: This tag:

.headerActionIconButtons > * {
 i { }
}

.headerActionIconButtons > * may apply only to the first children, but then the i {} cascades down and applies to all the children underneath that first-level child. So because these <i> tags exist somewhere underneath that direct descendant, they apply. Confusing I know.

I suggest you open up Header.jsx and moving this line a few lines down so it is after the /DIV that closes the .headerActionButton. Should be a really easy fix for this issue.

```
          {/* Render the SettingsDialog only when it is open */}
          {isDialogOpen && <SettingsDialog items={dashboardSettingsItems} className={'dashboard-settings'} onSaveChanges={handleChangesInSettings} />}

```

PR #709, opened 2025-12-29
Integrated Plugins - np.Shared, Forms & Favorites

dwertheimer:


jgclark:
Oof! 137 files changed. I'm not sensibly going to review all of that. But I will look at the shared stuff later, and leave any comments.

dwertheimer:
You really only need to look at HTMLView
Because your changes conflict with mine.

PR #710, opened 2025-12-29
Add GraphNote_V2 - interactive note graph visualisation

iamjameshannam:
Adds a new plugin: GraphNote_V2 (jh.GraphNote_V2)

- Interactive visualisation of note connections
- Multiple D3-based layouts (force, tree, radial, tangled, sunburst, icicle, treemap)
- Bundled dependency: d3.v7.min.js

Requesting review and release once merged.


jgclark:
Hi, James, and thanks for making this and sharing it. We're happy to help you get this published and available to everyone to install easily through the app.

First two comments: 
- Do you need to call it "GraphNote_V2" here? I'm sure you've had previous versions, but there's a separate semantic versioning system in use, and so I'd simplify it and call it "GraphNote", and pick either 1.0.0 or 2.0.0 as its version number.
- There's no source code here; that's not necessarily a problem, but its the only one without user-inspectable source.
- There's no `script.js` file either. This is a problem, as currently there's nothing to run.

PR #745, opened 2026-04-03
first release of the crm plugin

MarcoBarna:
This plugin is the first iteration in what i feels is good for noteplan meaning having the ability to Manage People as you manage projects and tasks


MarcoBarna:
I have noticed there are some issues on mobile so I need to fix them as I had a regression. Will do it on Tuesday

PR #750, opened 2026-06-08
Calendar plugin: support event colors stored in event notes

michalkrause:
Enable support for custom event colors saved within notes, providing compatibility with 'Calendars by Readdle' and 'Calendar 366' apps.

### Description
This PR adds support for parsing event colors stored inside the event notes. This behavior is used by some third-party calendar applications like 'Calendars by Readdle' and 'Calendar 366' to persist custom colors.

### Motivation
Users of these alternative calendar apps currently lose their color-coding when events are displayed in the Calendar plugin. This change improves ecosystem compatibility for those users.

### Risk Assessment & Implementation
While this addresses a non-standard feature specific to certain apps, the implementation is very straightforward and isolated:
- It safely falls back to default behavior if no color data is found in the note.
- It introduces no breaking changes or performance overhead for standard events.

### DIFF
--- a/emetzger.Calendar/script.js
+++ b/emetzger.Calendar/script.js
@@ -2178,6 +2178,16 @@ function getCalendarHTML() {
     }
 
     function getEventColor(event) {
+      // support for event color coding in Calendars by Readdle and Calendar 366 applications
+      if (event.notes) {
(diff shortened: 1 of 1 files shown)


jgclark:
Thank you Michael. I will update the documentation to cover this and then release.

michalkrause:
Thanks for accepting the patch!

PR #758, opened 2026-07-30
Dashboard v2.4.0.b58: reminder routing fallbacks + teamspace filter fix

dwertheimer:
@jgclark — flagging you as owner/reviewer. This PR shows the complete diff of everything after your `29d938ba`: five Dashboard fixes, plus the version bump to `2.4.0.b58` and changelog.

**One thing to be aware of:** equivalent commits (`84bddf54`…`16b08e91`) are *already* on `main` — they were pushed straight there during a live debugging session, before this PR existed, and `main` is force-push protected so they can't be withdrawn. This branch is rebuilt on `29d938ba` with those commits cherry-picked, so the diff you see below is the real, reviewable change set. The practical consequence: merging adds only the version bump and changelog as new content, since `main` already has the source changes. Review comments here still apply to live code — anything you want changed is a follow-up commit rather than a block.

All five were found by tracing a real symptom on a live install: *"why am I only seeing 3 reminders?"* Each was verified against running NotePlan, not just reasoned about.

## The fixes

| commit (this PR / on `main`) | what |
|---|---|
| `ce03cdb7` / `84bddf54` | Overdue counted reminders it didn't display |
| `54165422` / `8eec7ba0` | reminders whose section was off were dropped → fall back to REM |
| `d1453167` / `c4e88d80` | that fallback was resurrecting items hidden by *Show Current Reminders* |
| `24ac1f54` / `703dd712` | unreachable teamspace IDs hid **every task** |
| `fa8b2e3c` / `16b08e91` | untimed-today fallback + DESIGN DECISION note |

## Two worth your attention

**1. `24ac1f54` is the nasty one.** `includedTeamspaces` held six stale teamspace IDs and no `'private'`. Private notes are only read when `'private'` is in that list, so every note was filtered out before a single task could be read — reminders still showed, so it looked like tasks had vanished. It was also unfixable from the UI: signed out of Spaces, *Spaces to Include* correctly says "You are not a member of any Spaces" and offers nothing to click.

Both existing guards miss this — `?? ['private']` only fires when the setting is *absent*, and the `length === 0` check only when it's *empty*. A non-empty list of unreachable IDs sails through both. Now healed at runtime, so affected configs recover without editing settings.

**Still open, and not mine to fix:** something wrote those six IDs while zero Spaces were reachable, at the base `dashboardSettings` level. `MultiSelectSpaces.jsx` is the component showing "not a member of any Spaces" — if it can persist stale IDs while displaying nothing selectable, this will recur for anyone who signs out of Spaces.

**2. The design note in `fa8b2e3c`.** A timed reminder due later today is shown nowhere until its time arrives. That's your decision and I've left it as-is — but nothing recorded it, and from the outside it is *indistinguishable* from the silent-drop bug above. Separating the two took most of the session. The note names both places to change if you ever want them visible early, and flags that changing only one makes them appear twice.

## Verification

Same data throughout — buckets `timedToday=1 untimedToday=1 yesterday=0 tomorrow=0 overdue=4 undated=3` (sum 9):

| perspective | before | after |
|---|---|---|
| z_ALL_ON | 8 of 9 | 8 shown + 1 not-yet-due = **9** |
| Reminders Only - ALL | 7 of 9 | 8 shown + 1 not-yet-due = **9** |
| z_OVERDUE Only | 7 of 9 | 8 shown + 1 not-yet-due = **9** |
| Reminders Only - No Current | 7 of 9 | unchanged — correct, *Show Current Reminders* is off |

Every reminder is now either displayed or intentionally withheld. Tasks confirmed healthy after the teamspace fix (1 today + 7 overdue in `z_ALL_ON`). Dashboard tests **220/220**; no new Flow errors (4 in these files before and after).

## Two things to know before merging

- **Diagnostic logging is deliberately left in** at @dwertheimer's request for continued testing: one line per reminder (list, title, raw EventKit date vs derived date/time), one per bucket assignment, REM section size, and the names of reminder lists disabled in NotePlan. That last one resolved two "missing" reminders as simply living in lists NotePlan has switched off. Say the word and I'll strip them.
- **One non-bug worth knowing:** reminders only come from lists enabled in NotePlan. A blank `includedReminderLists` inherits NotePlan's enabled set; a non-empty CSV can select *any accessible* list, including ones NotePlan has switched off.


jgclark:
I'm confused. Somehow this has made it into my local code without merging this. 
So I'm closing without apparently merging it, but in practice **it is merged with thanks**, @dwertheimer

PR #767, opened 2026-08-02
Flow cleanup: 8,254 errors → 6, with a per-plugin CI ratchet

dwertheimer:
Repo-wide Flow cleanup. `flow check` goes from **8,254 errors to 6**, with no change to what any plugin does — except three deliberate bug fixes, each called out below.

@jgclark — the **"Over to you"** section at the bottom is the part that needs your decisions. Most of it is short.

---

## Numbers

| | Before | After |
|---|---|---|
| Raw `flow check` errors | 8,254 | **6** |
| Unique `file:line:code` sites | 1,314 | **6** |
| Files with errors | 205 | **3** |
| Plugins with any error | ~34 | **2 of 47** |
| Jest | 4,619 passing | **4,621 passing** |

The raw count and the site count started far apart because Flow emits one error per (source-type × destination-slot) pair — a single bad annotation could produce thousands of lines. One `reduce` accumulator in `dashboardSettingsClean.js` accounted for **3,160** of them on its own. The two numbers now coincide, which is itself evidence the cascades are gone.

Only `jgclark.PeriodicReviews` (4) and `np.Preview` (2) still error, and both are the same kind of problem: imports of modules or exports that do not exist. Neither is fixable with a type — see items 1 and 2.

---

## Did anything change at runtime?

This is the question worth checking, so there's a tool for it: `scripts/flow-emit-audit.js` Babel-transforms every changed file at a base revision and at the working tree, then diffs the *emitted JavaScript*. Flow annotations and casts are erased by `@babel/preset-flow`, so a genuinely type-only change must emit byte-identical output.

**179 of 205 changed files emitted byte-identical JavaScript.** The 26 exceptions:

| Category | Files | Note |
|---|---|---|
| Provably equivalent | 21 | `Number()` around a `Date` subtraction that `-` already coerced; empty statements from `;(x: any)` casts; `String()` where a template literal already called it; `const self = this` aliases; identity `asTNote()` wrappers |
| Verified equivalent, real edit | 1 | `sortTasks.js` `for...in` → indexed loop |
| Import corrected | 1 | `pluginToHTMLBridge.js` — a runtime const was sitting in an `import type` block |
| **Deliberate bug fixes** | 3 | below |

Plugin bundles rebuild unchanged (`jgclark.Dashboard`, `np.Templating`, `dwertheimer.TaskAutomations`, `.EventAutomations`, `.TaskSorting`, `.JestHelpers`, `np.CallbackURLs`, `jgclark.Reviews`, `np.plugin-test` all produce identical `script.js`).

You can re-run the audit yourself: `node scripts/flow-emit-audit.js` (add `--show` for the diffs).

### The three deliberate behaviour changes

1. **`helpers/dev.js`** — `logDebug`/`logInfo`/`logWarn`/`logError` are now variadic. Eleven call sites were passing a third argument that was being silently dropped, so **log output now includes data it previously discarded**. No other behaviour change.
2. **`dwertheimer.EventAutomations/src/NPTimeblocking.js`** — `writeSyncedCopies(...paras, config)` spread the array, so the function received the *first paragraph* as the array and the *second paragraph* as the config. The feature reported "No todos/references marked for this day!" and wrote nothing. A `// $FlowFixMe` sat directly above the line, suppressing the error that would have caught it.
3. **`dwertheimer.EventAutomations/src/config.js`** — `createSyncedCopies` existed in `plugin.json` but nowhere in `AutoTimeBlockingConfig`, its defaults, or `configTypeCheck`.

Three further incidental behaviour changes were caught by the audit and reverted to casts.

---

## What drove the bulk of the reduction

- **A `reduce` accumulator typed `Partial<TDashboardSettings>`** (exact, no indexer) for a dynamic-key copy, so Flow checked all ~85 possible value types against all ~85 destination slots. The same file's sibling `dashboardSettingsDefaults.js` already had the correct idiom.
- **`Partial<T>` property invariance.** A full `TDashboardSettings` is *not* assignable to `Partial<TDashboardSettings>`, because object properties are read-write. Added `TDashboardSettingsIn = $ReadOnly<Partial<TDashboardSettings>>` for the read-only paths — which is also the more honest type, since settings are assembled at runtime and Flow can't verify any "required" key is present.
- **`flow-typed/Noteplan.js` had real defects**: `values: object` (not a Flow type), an undeclared `TBacklinkNoteFields`, `$ReadOnlyArray<string, Date>` (two args to a one-arg type), two `[string]` indexers on `globalThis`, and `declare var Range` shadowing the DOM `Range` **class** — which meant Flow's own standard library stopped checking (12 errors in `dom.js`, now 0).
- **Test mocks.** ~5 hand-rolled `type MockNote = {...}` declarations, each detonating into 56–59 errors *per call site* under `exact_by_default`. Added `asTNote`/`asTParagraph` to `@mocks/index` and documented the pattern in `_README-Mocks.md`.

---

## Guard rail

`scripts/flow-report.js` collapses `flow check` output to unique `file:line:error-code` sites and compares against a committed baseline. A new `Flow-Check` CI job runs it with `--check`, failing only when a plugin's count *rises*. With 45 of 47 plugins at zero, the ratchet does real work.

Once items 1 and 2 are closed, this can be replaced with a plain `npx flow check`.

---

# Over to you, @jgclark

Nine items. Two are still erroring; seven are real defects that had to be silenced with a cast to satisfy the "no code changes" rule — those sites carry a `// KNOWN BUG` comment naming the actual problem, so Flow is quiet but the bug is not gone.

Every item below was re-verified at its stated line on 2026-08-02.

**Legend:** 🔴 user-visible breakage · 🟠 silent wrong behaviour · 🟡 dead code · 🔵 design question

### 1. `np.Preview` imports a function that no longer exists
`np.Preview/src/mathTests.js:4`, `np.Preview/src/mermaidTests.js:4`

```js
import { showHTML } from '@helpers/HTMLView'
```

`helpers/HTMLView.js` exports `showHTMLV2(body, opts)` and no `showHTML` at all. Both files call it positionally in ~8 places. They would throw on import.

**Suggestion:** port the call sites to `showHTMLV2(body, { windowTitle, ... })`, or delete both files if they were only ever scratch tests. Your call which.

### 2. `jgclark.PeriodicReviews/src` imports four modules that don't exist
`src/reviewHTMLViewGenerator.js:9, 10, 18, 19` — `../plugin.json`, `./periodicReviewHelpers` (×2), `./reviewQuestions`

That directory contains only this one file. It looks orphaned — left behind by a move or a partial extraction.

**Suggestion:** restore the missing files, or delete `src/`.

### 3. `KimMachineGun.Raindrop` opens `undefined` when the note already exists
`src/NPPluginMain.js:153-158`

`createNoteIfNotExists()` has no return on the "already exists" path, so it resolves to `undefined` and the caller does `Editor.openNoteByFilename(undefined)`.

**Suggestion:** `return existingNotes[0].filename` on that branch, then drop the cast.

### 4. `np.Tidy` — `runSilently` is never honoured
`src/tidyMain.js:46-47`

`removeOrphanedBlockIDs(params: string = '')` feeds `params` to `overrideSettingsWithEncodedTypedArgs()` and `getTagParamsFromString()`, but is passed the boolean `config.runSilently`. Its two neighbours in `tidyUpAll` genuinely do take booleans, which is presumably how it slipped in.

**Suggestion:** pass `'{"runSilently":true}'` (or `''`), matching the other params-taking calls below it.

### 5. `jgclark.Reviews` — a guard that can never be false
`src/projects.js:251` — `(finalCommentRaw && (finalCommentRaw: any) !== true)`

`getInputTrimmed()` returns `Promise<string | false>`, so `!== true` is always true.

**Suggestion:** drop the `&& ... !== true` clause.

### 6. `jgclark.Summaries` — vacuous string comparisons on numbers
`src/TMOccurrences.js` — 7 occurrences of `!== ''`

`this.count !== ''` / `this.total !== ''` where both are declared `number`. They filter nothing; the adjacent `!isNaN(...)` checks do the real work.

**Suggestion:** delete the `!== ''` clauses.

### 7. `NoTasks.jsx` references four undeclared globals
`jgclark.Dashboard/src/react/components/NoTasks.jsx` — 16 references

`random()`, `TweenLite`, `Power4` and `_` (lodash) are none of them imported or declared; line 10 has a commented-out `// import {random} from 'lodash'`. `ConfettiCannon` would throw `ReferenceError` if constructed. The file header says **"no longer used (from v2.2 or v2.3)"**.

**Suggestion:** delete the file. If you want the confetti later it needs GSAP as a dependency — a new decision, not a restoration.

### 8. `TEditor` is missing `linkedItems` and `datedTodos`
`flow-typed/Noteplan.js` — both are declared on `Note` (`:1736`, `:1743`) but not on `TEditor`

They appear to exist on the live Editor object; call sites work around it with `(e: any)` casts.

**Suggestion:** worth confirming with @EduardMe, then adding both to `TEditor` (or moving them to `CoreNoteFields`). Deliberately not done blind.

### 9. `TDashboardSettings` — 68 of 86 properties are "required" but nothing guarantees them
`jgclark.Dashboard/src/types.js`

The object is assembled at runtime by `getDashboardSettingsDefaults()` from a `TSettingItem` array, so Flow cannot verify any given key is present, and older installs won't have newer keys until defaults are merged. Consumers already treat them defensively (`dashboardSettings?.FFlag_UseTagCache !== false`).

**Suggestion:** audit the required/optional split — most probably want to be optional. `TDashboardSettingsIn` already handles the *input* side.

## Two lessons worth keeping

- **`flow check` passing is not enough.** Flow accepts syntax Babel cannot parse, and a file Babel can't parse breaks the build *and* jest. `flow-emit-audit.js` catches this.
- **Run the tests, not just Flow.** One regression during this work was invisible to `flow check` — a module that type-checked perfectly but threw `ReferenceError` on import, because it read NotePlan globals at module scope.

---


jgclark (approved):
Wow. Just wow. Well done, David and Claude.

PR #768, opened 2026-08-02
Remove 233 Flow suppressions that suppress nothing

dwertheimer:
Follow-up to #767 — **based on `flow-cleanup`, not `main`**, so review that one first. The diff here is 233 deleted comment lines and nothing else.

## What this removes

The repo carries **647** standalone `$FlowIgnore` / `$FlowFixMe` comments. **233 of them (36%) suppress nothing.** They're stale: the error they were written for got fixed later, or the comment drifted off the line it was covering. Thirteen of them sit in `scripts/rollup.js`, which has no `// @flow` pragma at all — so they never did anything.

| | |
|---|---|
| Files touched | 100 |
| Lines deleted | 233 (all comments) |
| Emitted JavaScript changed | **0 of 100 files** |
| Plugin bundles changed | 0 (`jgclark.Dashboard`, `np.Templating`, `dwertheimer.EventAutomations` rebuild identical) |
| `flow check` | 6 before → **6 after** |
| Jest | 4,621 passing, unchanged |

```
 49  np.Templating              15  dwertheimer.Forms       6  jgclark.Reviews
 46  jgclark.Dashboard          14  np.Shared               4  dwertheimer.TaskSorting
 36  helpers                     9  dwertheimer.Favorites   3  jgclark.MOCs
 16  dwertheimer.EventAutomations  6  dwertheimer.DateAutomations   3  np.Tidy
 16  scripts                                                       …and a tail
```

## How the list was produced

Worth reading, because the obvious method gives the wrong answer.

Parsing `flow check --include-suppressed` output *looks* like it identifies the dead suppressions — it reports which comment locations Flow used. Deleting the ones it didn't list took Flow from **6 errors to 38**. The attribution doesn't map cleanly onto comment lines, so that approach has false positives.

What actually works:

1. Replace **every** suppression comment with a same-length placeholder, so no line numbers shift.
2. Run `flow check` — 3,651 errors surface.
3. Restore only the placeholders that have an error on the line they cover — **414** of them.
4. Re-run — back to exactly **6** errors.

Anything still neutralised after step 3 is provably carrying no load. That's the 233 deleted here, and step 4 is the proof.

## Why it's safe to merge on sight

- Comment-only: `scripts/flow-emit-audit.js` confirms all 100 files emit byte-identical JavaScript.
- The `Flow-Check` CI ratchet from #767 will catch any regression, since it fails when a plugin's error count *rises*.
- If a suppression turns out to have been load-bearing after all, the symptom is a new Flow error in that file, not a runtime change.

## Not included

`$FlowIgnore`/`$FlowFixMe` comments that are **inline** or embedded in JSDoc prose (~130 lines) were left alone — deleting those would mean editing surrounding text rather than dropping a whole line. Only standalone comment lines are touched here.

The 414 that remain are load-bearing and stay. Reducing *those* means fixing the underlying types, which is a different piece of work.

---

### DIFF
--- a/dwertheimer.DateAutomations/src/dateFunctions.js
+++ b/dwertheimer.DateAutomations/src/dateFunctions.js
@@ -104,18 +104,15 @@ function getFormattedDateTime() {
           // conditionall add those keys to config
           if (ds !== '') {
             // Ignore type error for now
-            // $FlowFixMe
             config.dateStyle = ds
           }
           if (ts !== '') {
-            // $FlowFixMe
             config.timeStyle = ts
           }
           config.hour12 = h12
 
           const text = new Intl.DateTimeFormat(
             loc,
-            // $FlowFixMe
             config,
           ).format()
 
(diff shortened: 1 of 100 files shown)


dwertheimer:
Folded into #767 — the dead-suppression removal is comment-only and independently verified (100 files, byte-identical emitted JS, flow check unchanged at 6, tests green), so it does not need separate review. Keeping the PR chain shallow so the *real typing* work can land in its own PR where review attention is actually needed.

PR #769, opened 2026-08-02
Replace Flow suppressions with real types (560 → 18) and fix the bugs underneath

dwertheimer:
## What this PR is for

**Goal: stop hiding Flow errors and start fixing them.**

PR #767 got `flow check` from 8,254 errors down to 0 — but part of that was achieved by *suppressing* errors rather than resolving them, leaving 560 `$FlowIgnore` / `$FlowFixMe` comments across the repo. A suppression is worse than it looks: it silences **every** error on the following line, including ones introduced later, so it hides future bugs as well as the one it was written for.

This PR pays that debt back. It replaces suppressions with real type annotations wherever an honest type can be written, and — where a suppression turned out to be concealing an actual defect — fixes the defect.

**560 suppressions → 18** (15 of which are live; 3 are prose or commented-out code). `flow check` stays at 0. All 4,648 tests pass (4,644 existing plus 4 new).

The work also surfaced a number of genuine bugs — outright crashes, and values that were silently wrong rather than obviously so. Those are detailed below. It also found two dead commands: one had been throwing on every call for nine months and was deleted rather than repaired (#771).

### How to review this without reading 167 files

The important thing to know up front: **123 of the 167 changed files (74%) cannot possibly change behaviour**, and that is proven mechanically, not asserted. Babel erases Flow types at build time, so `scripts/flow-emit-audit.js` compiles every changed file before and after and diffs the emitted JavaScript — 119 files come out byte-identical, and 4 more differ only by a no-op `;`.

That leaves **44 files** where the risk genuinely lives. Section 3 breaks those down, and Section 5 lists the seven things worth testing by hand. **Section 3, Tier C is the short list of places where a human actually needs to make a decision** — there are five, and one of them is yours, @jgclark.

---

**Branch:** `flow-real-types` → `main` (rebased onto post-#767 main, now mergeable)
**Scope:** 167 files changed, 7 commits

---

## 1. Current state

| Metric | Before (#767 merged) | After #769 |
|---|---|---|
| `flow check` errors | 0 | **0** |
| Suppressions (`$FlowIgnore` / `$FlowFixMe`) | 560 | **18** |
| Jest | 200 suites / 4,644 | **201 suites / 4,648** |

**97% of remaining Flow suppressions removed.** Of the 18 left, 3 aren't real suppressions at all (two sit inside commented-out code; one is the literal text `$FlowIgnore[prop-missing]` inside a prose comment in `__mocks__/asNPTypes.js`). So **15 live suppressions remain**, each narrowed to a specific error code and carrying a comment explaining why it's still there.

The rebase dropped 6 commits that are now redundant (the multi-level indent fix and its docblock, both shipped in #770; the `insertTodos` revert and its correction, now obsolete; and a `[DIAG]` logging add/remove pair that nets to zero).

---

## 2. Where there is **zero** risk

This is the majority of the PR, and the claim is mechanical rather than a matter of judgement.

**119 of 167 files (71%) produce byte-identical JavaScript.**

Babel erases Flow types at build time, so a change that only touches annotations *cannot* alter runtime behaviour. `scripts/flow-emit-audit.js` proves this per-file: it runs the Babel transform on each changed file at `main` and again on this branch, and diffs the emitted JS. For 119 files the output is identical byte-for-byte.

**A further 4 files differ only by a bare `;`** — an empty statement, a no-op:

- `helpers/react/reactDev.js`
- `jgclark.Dashboard/src/react/components/Section/useSectionSortAndFilter.jsx`
- `np.Shared/src/react/Root.jsx`
- `np.Templating/__tests__/NPTemplateRunner.test.js`

The repo omits semicolons, so a statement starting with `(` continues the previous line. Inline casts at statement start must be written `;(foo: any).bar()`. That leading `;` survives type erasure as an empty statement.

**So 123 of 167 files carry no behavioural risk whatsoever, and this is verifiable by re-running the audit rather than by trusting review.**

---

## 3. Where the risk actually is

**44 files, 482 changed lines of emitted JavaScript.** I reviewed **100% of those lines at the emitted-JS level** — not the source diff. That matters: emitted JS has comments and type annotations stripped, so nothing can hide in a plausible-looking comment, and a "type-only" change that isn't shows up immediately.

The 44 files break into three tiers.

### Tier A — Mechanically equivalent refactors (no behaviour change)

Verified equivalent, usually by checking the thing the change depended on rather than assuming:

| File | Change | Why it's safe |
|---|---|---|
| `helpers/dev.js` | `timeEnd - timeStart` → `Number(timeEnd) - Number(timeStart)` | `Number()` is exactly what `-` does to each operand. **Important:** `.getTime()` would have been wrong — `helpers/promisePolyfill.js:161,183` pass `Date.now()` (a number) into a param annotated `Date`, so `.getTime()` would throw. |
| `helpers/note.js`, `helpers/NPnote.js` | 6 sorts → `Number(a) - Number(b)` | Same reasoning. An agent originally wrote `.getTime()` here; I reverted that because it adds a crash path the original `-` never had. |
| `dwertheimer.EventAutomations/src/config.js` | `validateConfigProperties` now gets a copy, result discarded | Verified it only *reads* `config[v]` — no writes anywhere in the function. |
| `dbludeau.TodoistNoteplanSync` | setters write `setup.x` not `this.x` | Verified `setup` is never spread or copied, so `this` was always `setup`. |
| `helpers/urls.js`, `jgclark.SearchExtensions/src/searchTriggers.js` | `while ((m = re.exec(s)))` restructured | Same `exec` call order and count. |
| `dwertheimer.TaskAutomations/src/NPTaskScanAndProcess.js` | `hasOwnProperty` → truthiness | Values are always non-empty arrays; `[]` is truthy anyway. |
| `dwertheimer.Forms/src/formSubmission.js` | `for…in` + `hasOwnProperty` → `Object.keys`; `Function.apply` → `new Function` | Identical semantics. |
| `jgclark.SearchExtensions/src/flexiSearch.js` | prefs wrapped in `String()` | Verified the values are only substituted into template strings, never used in boolean context (`"false"` would be truthy). |
| `helpers/config.js`, `helpers/NPConfiguration.js`, `helpers/content.js`, `dwertheimer.TaskSorting/src/tagTasks.js`, `np.Templating/lib/support/modules/data/service.js` | local refinements, `const` destructuring, explicit branches | Equivalent including null/array edge cases. |

### Tier B — Real bug fixes (behaviour changes, but the old behaviour was broken)

These *do* change behaviour. In each case the previous behaviour was a crash, an exception swallowed into a misleading message, or a silently wrong value.

**Crashes:**

- `helpers/NPParagraph.js` `paragraphMatches()` called `.startsWith()` on non-string fields (`lineIndex`, `indents`, `date`) → TypeError. Now type-guarded.
- `np.Tidy/src/tidyMain.js` `removeBlankNotes()` compared against the **string** `'undefined'`, so the guard was always true and a note with undefined content threw mid-filter.
- `jgclark.Dashboard` move-day/move-week handlers wrote `p.content` then used `p.note`, which really is `?TNote` — a genuine null dereference the suppression was hiding.
- `np.Templating/lib/support/modules/quote.js` resolved `undefined` out of a `Promise<string>` that callers assign to `const verse: string`.

**Silently wrong values:**

- `jgclark.Dashboard/src/reactMain.js` returned `{}` on error, so every setting read as `undefined`. Now returns `null` and the caller keeps the loaded settings.
- `helpers/NPWindows.js` `getStoredWindowRect()` returned a malformed preference as a fake `Rect`.
- `dwertheimer.Forms` `FormView` passed a tuple-taking `sendToPlugin` into context.
- `np.Templating/lib/support/modules/prompts/handlers/index.js` called `promptDateInterval('', message, default)` against a `(message, defaultValue)` signature — the user saw an empty prompt and the default was dropped.
- `helpers/NPDateStrings.js` — a `null` week threw, and the outer catch discarded **the entire relative-date list**. Now that one week is logged and skipped.

### Tier C — Judgement calls that need a human decision

**These are the only places where reasonable people could disagree.** Everything else is either provably inert or a fix to something demonstrably broken.

1. **`jgclark.SearchExtensions/src/searchHelpers.js` — `makeAnySyncs()` returns `input` instead of `null` on error.** *(@jgclark's call.)* The old behaviour wasn't a design: `null` escaped a `Promise<resultOutputType>` into `saveSearch()`, which reads `.resultCount` and threw — the whole search visibly failed. Now a failure while adding blockIDs costs only the sync links. The alternative (rethrow, so the user gets an empty result set) is equally defensible.

2. **Heading levels clamped to 1–8** (`helpers/userInput.js`, `helpers/paragraph.js`). A computed level of `0`, `NaN` or `12` previously went raw to `insertHeading()`; it's now rounded and clamped (`NaN` → 2). Only affects out-of-range input.

3. **`dwertheimer.TaskAutomations/src/NPOverdueReact.js` — field whitelist.** `para[field] = row[field]` became explicit `type`/`content` branches; anything else logs an error and is skipped. Verified only those two fields are ever sent, but a third would now be dropped rather than blind-assigned.

4. **`helpers/sorting.js` `fieldSorter` — two non-`Date` objects now compare equal.** I found this during review; the agent's claim that only strings/numbers reach that line was wrong. `isNaN()` coerces, so Dates and booleans still become numbers — only plain objects are affected, and the fields actually sorted are `content`/`index`/`date`/`priority`. Very unlikely to fire, but it *is* a divergence neither Flow nor the tests would catch.

5. **`dwertheimer.EventAutomations/src/events.js` — now skips untitled/undated events.** Worth knowing but **not testable**: its wrappers are exported but appear in neither `index.js` nor `plugin.json`, so the command is unreachable.

---

## 4. How we know this is safe

Four independent mechanisms, in decreasing order of strength:

1. **Emitted-JS audit (mechanical, re-runnable).** `node scripts/flow-emit-audit.js --base origin/main` — proves 119 files cannot behave differently. This is the single most important artifact in the PR: it converts "trust the review" into "check the tool."
2. **Full test suite.** 201 suites / 4,648 tests green. Note this is a *floor*, not a ceiling — coverage is thin in exactly the plugin code these changes touch.
3. **Line-by-line review of the entire behaviour-changing surface.** All 482 emitted lines across 44 files, not a sample. Six changes looked like defects and were chased individually; five turned out fine (and could only be known so by checking), one was a real divergence (Tier C item 4).
4. **Dead-suppression sweep.** Placeholder substitution — swap each suppression for a same-line-count placeholder so no line numbers shift, run Flow, see which covered lines actually error. Confirms no remaining suppression is inert (an inert suppression is worse than none: it hides future errors on that line).

### Things I got wrong during this work, and how they were caught

Stated plainly, because it calibrates how much to trust the rest:

- I claimed a Babel upgrade would unlock blocked syntax. **Wrong** — tested 4 versions, all fail. Corrected with a tested matrix.
- An agent's `chroma-js` change traded 3 errors for 5. Caught by re-measuring; reverted.
- A `checkType.js` mapped-type rewrite turned 19 suites red — Babel can't parse mapped types at any version. Reverted.
- Three incidental behaviour changes slipped into a "type-only" commit. Caught **by my own emit audit**, not by review.
- I twice attributed observed indent-flattening to a code path that never executed, and wrote the false conclusion into a commit message. Caught by @dwertheimer's logs; the fix that survived is the one he insisted on re-applying.

The pattern: **the mechanical checks caught what review missed.** That's why the audit matters more than my summary.

---

## 5. What a human should test before merging

Ordered by value. Items 1–2 are the ones I'd actually insist on.

**np.Templating is already verified** — @dwertheimer ran TemplateRunner and the meeting-note path end to end; the render pipeline (`templateProcessor.render`, `TemplatingEngine`, frontmatter parsing) came through clean. Its remaining changes are covered by unit tests, including a new static test for the `lib/rendering` barrel exports.

### 1. Dashboard settings survive a perspective switch
`Show Dashboard`, then switch perspectives several times including back to default.
✅ Settings persist. ❌ Settings blank out, or the window fails to open. Check the log for `carry on with the existing settings` or `Couldn't assemble the initial data`.
*Why:* `reactMain` used to return `{}` on error, silently replacing every setting with `undefined`.

### 2. Search with sync'd results (jgclark)
Settings: result style **NotePlan**, `syncOpenResultItems` **on**. Run `search` for a term matching several open tasks.
✅ Results carry sync IDs (`^abc123`) exactly as before. The happy path must be untouched — that's the main thing to confirm; the changed path only fires on an error that already broke the search.

### 3. Relative-date lists
Any chooser showing relative dates (e.g. `Show Dashboard`, or a template using them).
✅ `this week`, `last week`, `next week`, `2 weeks ago` … `10 weeks' time` all present. ❌ Whole categories missing → check log for `returned null for week offset`.

### 4. Window position restore
Open a plugin window, move/resize, close, reopen. ✅ Restored. If the log shows `isn't a valid Rect`, you had a corrupt stored pref — the intended fix, but worth knowing it fired.

### 5. Rename a note
`rename note` on a normal note ✅; on a title-less note ✅ stops cleanly rather than offering `"undefined"`.

**`dwertheimer.TaskAutomations` overdue React window — not testable, and not because of this PR.** The `paragraphUpdateReceived` field-whitelist change lives in that window's message router. The window does not render at all: its bundle throws `TypeError: i.default.div is not a function` before defining the `WebView` global, so `Root` then fails with `ReferenceError: WebView is not defined`. Diagnosed as a rollup ESM→CJS interop fault — `getAugmentedNamespace()` wraps styled-components' function default export in a *function*, which defeats react-data-table-component's `typeof === 'object'` interop check and double-wraps it, so `.div` is missing. **Pre-existing and unrelated to this PR**, which does not touch that plugin's React source at all (verified: the emitted JS of every file under `src/react/` is unchanged). The commands that open it are all `hidden: true` and the view is not in use, so it is not being fixed here.

**Not testable — dead or unreachable code, no action needed:** `events.js` (not referenced by `index.js` or `plugin.json`), `TemplatingEngine.getDefaultFormat()` (marked "should never be called"). The `renderTemplate` barrel re-export has no importer either, so it is pinned by a static test (`__tests__/rendering-barrel-exports.test.js`) rather than by hand — that test asserts *no* named export of the barrel resolves to `undefined`, which catches this class of bug generically.

---

## 6. Known pre-existing bugs found but deliberately NOT fixed

Listed so they aren't mistaken for regressions during testing:

- **`helpers/calendar.js` `keepTodayPortionOnly()`** claims `Array<TCalendarItem>` but returns hand-built plain objects with no `id`, no `isRecurring`, no methods. Any caller reading `.id` or calling `.findLinkedFilenames()` on a multi-day event gets `undefined` or a crash. Four files across three plugins consume it, and `helpers/__tests__/calendar.test.js` asserts `toEqual()` on the plain-object shape — so the test currently locks in the wrong contract.
- **`helpers/NPnote.js` `chooseNoteV2()`** can return a fake note: pick a future calendar date with no note yet and you get a stub with only `title`/`type`/`changedDate`.
- **`helpers/urls.js` `findProjectNoteUrlInText()`** — the `allowCalendarNotes` parameter is dead; the filtered result is computed then discarded.
- **`jgclark.SearchExtensions/flexiSearch.js`** — `String(x) ?? 'quick'` dead defaults: `String()` never returns null, so unset preferences become the literal string `"undefined"`.
- **`dbludeau.TodoistNoteplanSync`** — `teamAccount` is both a data property and a setter; the setter wins, so reads return `undefined`. Appears to be dead code.
- **`helpers/sorting.js` `fieldSorter`** — declared `(a: string, b: string)` but always called with objects; also compares against the *string* `'NaN'`.

---

## 7. Separate finding: NotePlan API bug (not ours, not fixable by plugins)

**Space-indentation is invisible to plugins.** Verified against a note whose bytes on disk were confirmed:

| Line | On disk | API `indents` | API `rawContent` |
|---|---|---|---|
| tab child | 1 tab | **1** ✓ | `"\t* A child ONE-TAB"` ✓ |
| tab grandchild | 2 tabs | **2** ✓ | tabs preserved ✓ |
| 2-space child | 2 spaces | **0** ✗ | spaces stripped |
| 8-space grandchild | 8 spaces | **0** ✗ | spaces stripped |

`rawContent` is reconstructed from marker + `indents` rather than being the literal file bytes (a `- [ ]` on disk comes back as `* `), so once the parser declines to count leading spaces the indentation is gone from *both* fields before any plugin runs.

**The consequence is data loss, not display:** any plugin that deletes and re-inserts paragraphs writes back the stripped version and permanently flattens a space-indented file. Repro note is `DELETEME/indent-api-test.md`; section D shows both behaviours under a single parent. Needs raising with @EduardMe.


dwertheimer:
## Smoke-test checklist for the behaviour changes in this PR

Of the 71 changed files, **25 emit byte-identical JavaScript** (provably type-only — nothing to test). The other 46 were reviewed line-by-line at the emitted-JS level, which yielded exactly **11 behaviour changes**. This checklist covers all 11. Nothing else in the diff changes runtime behaviour.

### Before you start

Build the plugins you're testing, then keep a log tail running:

```bash
npc plugin:dev <plugin-id> -nc
node scripts/nplog/nplog --json --follow --wait-idle 5
```

Do **not** grep the raw log file — the three obvious greps all silently return zero results. Use `nplog`.

A ❌ on any check means: tell me which one, and I'll revert that specific change back to a suppression.

---

### 1. `insertTodos` now passes `'text'` to the NotePlan API — **highest risk**

*Was:* `note.insertParagraphAfterParagraph(content, lastPara)` — `paragraphType` arrived `undefined`.
*Now:* `note.insertParagraphAfterParagraph(content, lastPara, 'text')`.

This is the one neither Flow nor jest can validate, because it depends on how the native bridge treats a missing argument.

1. Make a note with a heading, several open tasks under it, and **no separator (`---`) as the last line** of that block.
2. Run **`Sort tasks under heading (choose)`** and pick that heading.
3. ✅ **Pass:** tasks reappear sorted, each rendering as a normal task (`- [ ]`), indentation preserved, no duplicated or missing `- [ ]` markers.
4. ❌ **Fail:** lines render as plain text, markers doubled (`- [ ] - [ ]`), or content lands with the wrong type.
5. Repeat with a block whose **last line *is* a `---` separator** — that path calls `insertTodoBeforeParagraph` and is unchanged, so it's a control.

### 2. `makeAnySyncs()` returns the unsynced result set instead of `null` on error

*Was:* returned `null`, which crashed `saveSearch()` with a `TypeError` — the whole search visibly failed.
*Now:* returns the results without sync links.

**@jgclark — this is the policy call flagged in the description.** Both recoveries are defensible.

1. Settings: result style **NotePlan**, `syncOpenResultItems` **on**.
2. Run **`search`** for a term matching several open tasks.
3. ✅ **Pass:** results appear with sync'd line IDs (`^abc123`) as before. No change in the happy path.
4. ❌ **Fail:** results missing sync IDs when they used to have them.

### 3. Overdue-review edits are restricted to `type` and `content`

*Was:* `para[field] = row[field]` — blind assignment of any field.
*Now:* only `type` and `content`; anything else logs an error and is skipped.

1. Run **`Review overdue tasks (by Task)`**.
2. Edit a task's **text** and save → ✅ text updates.
3. **Complete / cancel** a task from the dialog → ✅ type changes and it leaves the overdue list.
4. Check `nplog` for `unsupported field` — ✅ **Pass:** absent. ❌ **Fail:** present (tells us a third field was in use; send me the name).

### 4. Meeting-note templating previously threw every time

*Was:* `NPTemplating.getTemplate()` — **does not exist**; threw a `TypeError` on that line.
*Now:* `getTemplateContent()`.

1. Run **`Create Meeting Note using Meeting Note Template`** and pick a template.
2. ✅ **Pass:** the template renders into the note.
3. ❌ **Fail:** anything worse than before. (Note: this path was *always* broken, so "it now works" is the expected outcome — if it still fails, the failure is downstream of what I fixed.)

### 5. Window position/size restore rejects malformed preferences

*Was:* a partial/corrupt stored pref was returned as a fake `Rect` and used.
*Now:* logs a warning and behaves as if no pref were stored.

1. Open a plugin window (e.g. **`Show Dashboard`**), move and resize it, close it, reopen.
2. ✅ **Pass:** position and size restored as before.
3. Check `nplog` for `isn't a valid Rect` — if it appears, you had a corrupt pref and the window will open at default size. That's the intended fix, but tell me if it fires, because it means real stored data is malformed.

### 6. Dashboard settings no longer silently blanked on error

*Was:* returned `{}` — every setting read as `undefined`.
*Now:* returns `null`; the caller keeps existing settings and logs a warning.

1. **`Show Dashboard`** → ✅ opens with your settings intact.
2. Switch perspectives a few times, including back to the default.
3. ✅ **Pass:** settings persist across switches; sections/filters as expected.
4. ❌ **Fail:** settings reset to blank/defaults, or the window fails to open. Check `nplog` for `carry on with the existing settings` or `Couldn't assemble the initial data`.

### 7. Relative-date lists skip an unresolvable week instead of aborting

*Was:* a `null` from `getNPWeekData()` threw, and the outer catch discarded **the entire list** (returning `[{}]`).
*Now:* that one week is logged and skipped; every other entry is still returned.

1. Trigger any note chooser showing relative dates (e.g. **`Show Dashboard`**, or a template using relative dates).
2. ✅ **Pass:** you see `this week`, `last week`, `next week`, `2 weeks ago` … through `10 weeks' time`, plus days/months/quarters.
3. ❌ **Fail:** whole categories missing. Check `nplog` for `returned null for week offset`.

### 8. Heading levels are clamped to 1–8

*Was:* a computed level of 0, 9, or `NaN` went raw to `insertHeading()`.
*Now:* rounded and clamped to 1–8 (`NaN` → 2).

1. Use a command that inserts headings at a computed depth — e.g. filing into a nested heading, or a template that creates sections.
2. ✅ **Pass:** headings render at sensible depths (`#`–`########`).
3. ❌ **Fail:** a heading that used to render at one depth now renders at another.

### 9. `renameNotes` guard widened from `title === ''` to `!title`

1. Run **`rename note`** on a normal note → ✅ renames as before.
2. Run it on a note with no title → ✅ stops cleanly with "No title found" rather than offering `"undefined"`.

### 10. `fieldSorter` treats two non-`Date` objects as equal

*Was:* compared them with `>` (deterministic but arbitrary). *Now:* returns `0`.

I found this one during review; the agent's claim that only strings/numbers reach that line was wrong. `isNaN()` coerces, so Dates and booleans still become numbers — only plain objects are affected, and the fields actually sorted are `content`/`index`/`date`/`priority`. **Very unlikely to fire.**

1. Run **`Tasks Sort by User Default`** and a couple of the other sorts on a busy note.
2. ✅ **Pass:** ordering matches what you'd expect / what it did before.

### 11. `quote()` returns a message instead of `undefined`

*Was:* resolved `undefined` out of a `Promise<string>` when the service returned an empty array.
*Now:* returns `**quote() web service returned no quotes**`.

1. Render a template using `quote()`.
2. ✅ **Pass:** a quote appears, or a readable error message — never the literal `undefined`.

---

### Not testable — dead code (no action needed)

- **`events.js`** (`createNoteForCalendarItem`): its wrappers are exported but appear in **neither `index.js` nor `plugin.json`**, so the command is unreachable. The null-guard and untitled/undated-event skip it gained cannot be triggered.
- **`TemplatingEngine.getDefaultFormat()`**: marked `FIXME` / "should never be called"; it called a non-existent method and always threw.
- **`renderTemplate` barrel re-export**: exported `undefined`; nothing in the repo imports it.

### Pre-existing bugs found but deliberately **not** fixed

These are unchanged by this PR — listed so they aren't mistaken for regressions during testing: `helpers/calendar.js` `keepTodayPortionOnly()` returning fake `TCalendarItem`s; `chooseNoteV2()` returning a stub note for future calendar dates; `findProjectNoteUrlInText()`'s dead `allowCalendarNotes` param; `flexiSearch`'s `String(x) ?? 'quick'` dead defaults; dbludeau's duplicate `teamAccount` key.

dwertheimer:
# PR #769 — Status, Risk Analysis, and Test Plan

**Branch:** `flow-real-types` → `main` (rebased onto post-#767 main, now mergeable)
**Scope:** 167 files changed, 7 commits

---

## 1. Current state

| Metric | Before (#767 merged) | After #769 |
|---|---|---|
| `flow check` errors | 0 | **0** |
| Suppressions (`$FlowIgnore` / `$FlowFixMe`) | 560 | **18** |
| Jest | 200 suites / 4,644 | **200 suites / 4,644** |

**97% of remaining Flow suppressions removed.** Of the 18 left, 3 aren't real suppressions at all (two sit inside commented-out code; one is the literal text `$FlowIgnore[prop-missing]` inside a prose comment in `__mocks__/asNPTypes.js`). So **15 live suppressions remain**, each narrowed to a specific error code and carrying a comment explaining why it's still there.

The rebase dropped 6 commits that are now redundant (the multi-level indent fix and its docblock, both shipped in #770; the `insertTodos` revert and its correction, now obsolete; and a `[DIAG]` logging add/remove pair that nets to zero).

---

## 2. Where there is **zero** risk

This is the majority of the PR, and the claim is mechanical rather than a matter of judgement.

**118 of 167 files (71%) produce byte-identical JavaScript.**

Babel erases Flow types at build time, so a change that only touches annotations *cannot* alter runtime behaviour. `scripts/flow-emit-audit.js` proves this per-file: it runs the Babel transform on each changed file at `main` and again on this branch, and diffs the emitted JS. For 118 files the output is identical byte-for-byte.

**A further 4 files differ only by a bare `;`** — an empty statement, a no-op:

- `helpers/react/reactDev.js`
- `jgclark.Dashboard/src/react/components/Section/useSectionSortAndFilter.jsx`
- `np.Shared/src/react/Root.jsx`
- `np.Templating/__tests__/NPTemplateRunner.test.js`

The repo omits semicolons, so a statement starting with `(` continues the previous line. Inline casts at statement start must be written `;(foo: any).bar()`. That leading `;` survives type erasure as an empty statement.

**So 122 of 167 files carry no behavioural risk whatsoever, and this is verifiable by re-running the audit rather than by trusting review.**

---

## 3. Where the risk actually is

**45 files, 489 changed lines of emitted JavaScript.** I reviewed **100% of those lines at the emitted-JS level** — not the source diff. That matters: emitted JS has comments and type annotations stripped, so nothing can hide in a plausible-looking comment, and a "type-only" change that isn't shows up immediately.

The 45 files break into three tiers.

### Tier A — Mechanically equivalent refactors (no behaviour change)

Verified equivalent, usually by checking the thing the change depended on rather than assuming:

| File | Change | Why it's safe |
|---|---|---|
| `helpers/dev.js` | `timeEnd - timeStart` → `Number(timeEnd) - Number(timeStart)` | `Number()` is exactly what `-` does to each operand. **Important:** `.getTime()` would have been wrong — `helpers/promisePolyfill.js:161,183` pass `Date.now()` (a number) into a param annotated `Date`, so `.getTime()` would throw. |
| `helpers/note.js`, `helpers/NPnote.js` | 6 sorts → `Number(a) - Number(b)` | Same reasoning. An agent originally wrote `.getTime()` here; I reverted that because it adds a crash path the original `-` never had. |
| `dwertheimer.EventAutomations/src/config.js` | `validateConfigProperties` now gets a copy, result discarded | Verified it only *reads* `config[v]` — no writes anywhere in the function. |
| `dbludeau.TodoistNoteplanSync` | setters write `setup.x` not `this.x` | Verified `setup` is never spread or copied, so `this` was always `setup`. |
| `helpers/urls.js`, `jgclark.SearchExtensions/src/searchTriggers.js` | `while ((m = re.exec(s)))` restructured | Same `exec` call order and count. |
| `dwertheimer.TaskAutomations/src/NPTaskScanAndProcess.js` | `hasOwnProperty` → truthiness | Values are always non-empty arrays; `[]` is truthy anyway. |
| `dwertheimer.Forms/src/formSubmission.js` | `for…in` + `hasOwnProperty` → `Object.keys`; `Function.apply` → `new Function` | Identical semantics. |
| `jgclark.SearchExtensions/src/flexiSearch.js` | prefs wrapped in `String()` | Verified the values are only substituted into template strings, never used in boolean context (`"false"` would be truthy). |
| `helpers/config.js`, `helpers/NPConfiguration.js`, `helpers/content.js`, `dwertheimer.TaskSorting/src/tagTasks.js`, `np.Templating/lib/support/modules/data/service.js` | local refinements, `const` destructuring, explicit branches | Equivalent including null/array edge cases. |

### Tier B — Real bug fixes (behaviour changes, but the old behaviour was broken)

These *do* change behaviour. In each case the previous behaviour was a crash, an exception swallowed into a misleading message, or a silently wrong value.

**Three functions that did not exist** — confirmed absent by grep, not inferred:

| Call site | Was | Now |
|---|---|---|
| `np.Templating/src/Templating.js` | `NPTemplating.getTemplate()` — **threw a TypeError every time**, so meeting-note templating was broken | `getTemplateContent()` |
| `np.Templating/lib/TemplatingEngine.js` | `this.getTemplateConfig()` — no such method | `this.templateConfig` |
| `np.Templating/lib/rendering/index.js` | re-exported `renderTemplate` — resolved to `undefined` | `renderTemplateByName` |

**Crashes:**

- `helpers/NPParagraph.js` `paragraphMatches()` called `.startsWith()` on non-string fields (`lineIndex`, `indents`, `date`) → TypeError. Now type-guarded.
- `np.Tidy/src/tidyMain.js` `removeBlankNotes()` compared against the **string** `'undefined'`, so the guard was always true and a note with undefined content threw mid-filter.
- `jgclark.Dashboard` move-day/move-week handlers wrote `p.content` then used `p.note`, which really is `?TNote` — a genuine null dereference the suppression was hiding.
- `np.Templating/lib/support/modules/quote.js` resolved `undefined` out of a `Promise<string>` that callers assign to `const verse: string`.

**Silently wrong values:**

- `jgclark.Dashboard/src/reactMain.js` returned `{}` on error, so every setting read as `undefined`. Now returns `null` and the caller keeps the loaded settings.
- `helpers/NPWindows.js` `getStoredWindowRect()` returned a malformed preference as a fake `Rect`.
- `dwertheimer.Forms` `FormView` passed a tuple-taking `sendToPlugin` into context.
- `np.Templating/lib/support/modules/prompts/handlers/index.js` called `promptDateInterval('', message, default)` against a `(message, defaultValue)` signature — the user saw an empty prompt and the default was dropped.
- `helpers/NPDateStrings.js` — a `null` week threw, and the outer catch discarded **the entire relative-date list**. Now that one week is logged and skipped.

### Tier C — Judgement calls that need a human decision

**These are the only places where reasonable people could disagree.** Everything else is either provably inert or a fix to something demonstrably broken.

1. **`jgclark.SearchExtensions/src/searchHelpers.js` — `makeAnySyncs()` returns `input` instead of `null` on error.** *(@jgclark's call.)* The old behaviour wasn't a design: `null` escaped a `Promise<resultOutputType>` into `saveSearch()`, which reads `.resultCount` and threw — the whole search visibly failed. Now a failure while adding blockIDs costs only the sync links. The alternative (rethrow, so the user gets an empty result set) is equally defensible.

2. **Heading levels clamped to 1–8** (`helpers/userInput.js`, `helpers/paragraph.js`). A computed level of `0`, `NaN` or `12` previously went raw to `insertHeading()`; it's now rounded and clamped (`NaN` → 2). Only affects out-of-range input.

3. **`dwertheimer.TaskAutomations/src/NPOverdueReact.js` — field whitelist.** `para[field] = row[field]` became explicit `type`/`content` branches; anything else logs an error and is skipped. Verified only those two fields are ever sent, but a third would now be dropped rather than blind-assigned.

4. **`helpers/sorting.js` `fieldSorter` — two non-`Date` objects now compare equal.** I found this during review; the agent's claim that only strings/numbers reach that line was wrong. `isNaN()` coerces, so Dates and booleans still become numbers — only plain objects are affected, and the fields actually sorted are `content`/`index`/`date`/`priority`. Very unlikely to fire, but it *is* a divergence neither Flow nor the tests would catch.

5. **`dwertheimer.EventAutomations/src/events.js` — now skips untitled/undated events.** Worth knowing but **not testable**: its wrappers are exported but appear in neither `index.js` nor `plugin.json`, so the command is unreachable.

---

## 4. How we know this is safe

Four independent mechanisms, in decreasing order of strength:

1. **Emitted-JS audit (mechanical, re-runnable).** `node scripts/flow-emit-audit.js --base origin/main` — proves 118 files cannot behave differently. This is the single most important artifact in the PR: it converts "trust the review" into "check the tool."
2. **Full test suite.** 200 suites / 4,644 tests green. Note this is a *floor*, not a ceiling — coverage is thin in exactly the plugin code these changes touch.
3. **Line-by-line review of the entire behaviour-changing surface.** All 489 emitted lines across 45 files, not a sample. Six changes looked like defects and were chased individually; five turned out fine (and could only be known so by checking), one was a real divergence (Tier C item 4).
4. **Dead-suppression sweep.** Placeholder substitution — swap each suppression for a same-line-count placeholder so no line numbers shift, run Flow, see which covered lines actually error. Confirms no remaining suppression is inert (an inert suppression is worse than none: it hides future errors on that line).

### Things I got wrong during this work, and how they were caught

Stated plainly, because it calibrates how much to trust the rest:

- I claimed a Babel upgrade would unlock blocked syntax. **Wrong** — tested 4 versions, all fail. Corrected with a tested matrix.
- An agent's `chroma-js` change traded 3 errors for 5. Caught by re-measuring; reverted.
- A `checkType.js` mapped-type rewrite turned 19 suites red — Babel can't parse mapped types at any version. Reverted.
- Three incidental behaviour changes slipped into a "type-only" commit. Caught **by my own emit audit**, not by review.
- I twice attributed observed indent-flattening to a code path that never executed, and wrote the false conclusion into a commit message. Caught by @dwertheimer's logs; the fix that survived is the one he insisted on re-applying.

The pattern: **the mechanical checks caught what review missed.** That's why the audit matters more than my summary.

---

## 5. What a human should test before merging

Ordered by value. Items 1–2 are the ones I'd actually insist on.

### 1. Meeting notes (np.Templating) — **highest value**
`Create Meeting Note using Meeting Note Template`.
This path **threw a TypeError on every invocation** before this PR (`NPTemplating.getTemplate` doesn't exist). Expect it to work now. If it still fails, the failure is downstream of what was fixed.

### 2. Dashboard settings survive a perspective switch
`Show Dashboard`, then switch perspectives several times including back to default.
✅ Settings persist. ❌ Settings blank out, or the window fails to open. Check the log for `carry on with the existing settings` or `Couldn't assemble the initial data`.
*Why:* `reactMain` used to return `{}` on error, silently replacing every setting with `undefined`.

### 3. Search with sync'd results (jgclark)
Settings: result style **NotePlan**, `syncOpenResultItems` **on**. Run `search` for a term matching several open tasks.
✅ Results carry sync IDs (`^abc123`) exactly as before. The happy path must be untouched — that's the main thing to confirm; the changed path only fires on an error that already broke the search.

### 4. Overdue review edits
`Review overdue tasks (by Task)`. Edit a task's text; complete/cancel a task.
✅ Both work. Check log for `unsupported field` — its absence confirms only `type`/`content` are ever sent.

### 5. Relative-date lists
Any chooser showing relative dates (e.g. `Show Dashboard`, or a template using them).
✅ `this week`, `last week`, `next week`, `2 weeks ago` … `10 weeks' time` all present. ❌ Whole categories missing → check log for `returned null for week offset`.

### 6. Window position restore
Open a plugin window, move/resize, close, reopen. ✅ Restored. If the log shows `isn't a valid Rect`, you had a corrupt stored pref — the intended fix, but worth knowing it fired.

### 7. Rename a note
`rename note` on a normal note ✅; on a title-less note ✅ stops cleanly rather than offering `"undefined"`.

**Not testable — dead code, no action needed:** `events.js` (unreachable — not in `index.js` or `plugin.json`), `TemplatingEngine.getDefaultFormat()` (marked "should never be called"), the `renderTemplate` barrel re-export (nothing imports it).

---

## 6. Known pre-existing bugs found but deliberately NOT fixed

Listed so they aren't mistaken for regressions during testing:

- **`helpers/calendar.js` `keepTodayPortionOnly()`** claims `Array<TCalendarItem>` but returns hand-built plain objects with no `id`, no `isRecurring`, no methods. Any caller reading `.id` or calling `.findLinkedFilenames()` on a multi-day event gets `undefined` or a crash. Four files across three plugins consume it, and `helpers/__tests__/calendar.test.js` asserts `toEqual()` on the plain-object shape — so the test currently locks in the wrong contract.
- **`helpers/NPnote.js` `chooseNoteV2()`** can return a fake note: pick a future calendar date with no note yet and you get a stub with only `title`/`type`/`changedDate`.
- **`helpers/urls.js` `findProjectNoteUrlInText()`** — the `allowCalendarNotes` parameter is dead; the filtered result is computed then discarded.
- **`jgclark.SearchExtensions/flexiSearch.js`** — `String(x) ?? 'quick'` dead defaults: `String()` never returns null, so unset preferences become the literal string `"undefined"`.
- **`dbludeau.TodoistNoteplanSync`** — `teamAccount` is both a data property and a setter; the setter wins, so reads return `undefined`. Appears to be dead code.
- **`helpers/sorting.js` `fieldSorter`** — declared `(a: string, b: string)` but always called with objects; also compares against the *string* `'NaN'`.

---

## 7. Separate finding: NotePlan API bug (not ours, not fixable by plugins)

**Space-indentation is invisible to plugins.** Verified against a note whose bytes on disk were confirmed:

| Line | On disk | API `indents` | API `rawContent` |
|---|---|---|---|
| tab child | 1 tab | **1** ✓ | `"\t* A child ONE-TAB"` ✓ |
| tab grandchild | 2 tabs | **2** ✓ | tabs preserved ✓ |
| 2-space child | 2 spaces | **0** ✗ | spaces stripped |
| 8-space grandchild | 8 spaces | **0** ✗ | spaces stripped |

`rawContent` is reconstructed from marker + `indents` rather than being the literal file bytes (a `- [ ]` on disk comes back as `* `), so once the parser declines to count leading spaces the indentation is gone from *both* fields before any plugin runs.

**The consequence is data loss, not display:** any plugin that deletes and re-inserts paragraphs writes back the stripped version and permanently flattens a space-indented file. Repro note is `DELETEME/indent-api-test.md`; section D shows both behaviours under a single parent. Needs raising with @EduardMe.

jgclark:
On "Search with sync'd results (jgclark)" I'm not going to fix. This is on old 2.x work, and I've since done a large amount in a new 3.x branch that this has not reviewed.

On "Window Resize" this has uncovered a range of problems stopping saving-on-resize-or-move in Dashboard and Project Review Lists.  These I'm fixing.

dwertheimer:
@jgclark As you wish...fix whatever you think is worth fixing. I'm inclined to merge it to have things be clean moving forward. We will need to tell all the contributors to re-pull their code.

jgclark:
I've now checked item 2 (Dashboard Perspectives) and that seems fine.  So I will now merge.

PR #770, opened 2026-08-03
TaskSorting 1.3.0: fix multi-level indented subtasks being flattened when sorting

dwertheimer:
Standalone fix against `main`, so TaskSorting can be released independently of the Flow work in #767/#769.

## The bug

Sorting flattened nested subtasks. `getTasksByType()` tracked a single `lastParent`, which was only reassigned when a **top-level** task was seen — so anything indented more than once compared its indent against the top-level task and got pushed onto *that* task's `children` array.

A grandchild therefore came back out as **a sibling of its own parent**, and every level below the first collapsed into one:

```
- [ ] apple                 - [ ] apple
	- [ ] child        ->      	- [ ] child
		- [ ] grandchild        	- [ ] grandchild   <- promoted a level
```

## The fix

An ancestor stack that unwinds to the correct parent, so nesting is tracked to arbitrary depth, and a sorted task carries its whole subtree with it at the original depths.

Two subtleties, both caught by tests rather than by reading the code:

1. **Unwind before classifying the paragraph, not after.** Unwinding afterwards compares a line against its own preceding *sibling*, which silently drops non-task lines sitting at sibling depth. The existing `taskDocument.json` fixture has exactly this — `\t> A quote after Subtask1-2` is at the same indent as the subtasks around it — and it vanished entirely.
2. **Treat a missing `indents` as 0.** NotePlan always sets it, but hand-built paragraphs (tests, other plugins) often don't, and `undefined <= undefined` is `false`, so the stack never unwound and every task nested under the first one. The old `{indents: 999}` sentinel had been tolerating this by accident. This broke 10 existing EventAutomations/TaskSorting tests until fixed.

## The duplicate-lines bug this also fixes

All five consumers in `sortTasks.js` did a one-level `children.forEach()`. **Two of them collect paragraphs for deletion** before re-inserting in sorted order. With real nesting, those would have missed grandchildren — deleting the parent and child, re-inserting all three, and **leaving duplicates in the note**. They now walk the whole subtree via a new `getAllDescendants()`.

## Deliberately not done: sorting subtasks

Children keep their **document order**; only the top level is re-ordered.

A parent's children are not all tasks. Notes and quotes indented at sibling depth are ordinary children, and their meaning is positional — the fixture's `> A quote after Subtask1-2` is exactly this, and alphabetical sorting moves it away from the line it names. `sortTaskTree()` is implemented, exported and tested for callers with all-task subtrees, but is not wired up by default; its docblock records why.

## Tests

New `helpers/__tests__/sorting-indents.test.js` — 18 tests: tree building, four-level nesting, dedent handling, non-task children, `ignoreIndents`, descendant ordering, and a round trip asserting that **every source line appears exactly once** and that **the deletion set matches the re-insert set** (the invariant that, if broken, duplicates or drops lines in the real note).

I verified the tests actually discriminate rather than just passing: replaying the old algorithm on the grandchild case puts it in `parent.children` alongside `child`, which these tests reject.

Full suite: **201 suites / 4,637 tests** green.

## Known NotePlan limitation found while testing this

Indentation made of **spaces is invisible to plugins**. Verified against a note whose bytes on disk were confirmed:

| line | on disk | API `indents` | API `rawContent` |
|---|---|---|---|
| tab child | 1 tab | **1** ✓ | `"\t* A child ONE-TAB"` ✓ |
| tab grandchild | 2 tabs | **2** ✓ | tabs preserved ✓ |
| 2-space child | 2 spaces | **0** ✗ | spaces stripped |
| 8-space grandchild | 8 spaces | **0** ✗ | spaces stripped |

`rawContent` is reconstructed from marker + `indents` rather than being the literal file bytes (a `- [ ]` on disk comes back as `* `), so once the parser declines to count leading spaces, the indentation is gone from *both* fields before any plugin runs. **No plugin-side workaround is possible.**

The consequence is data loss, not display: any plugin that deletes and re-inserts paragraphs writes back the stripped version and permanently flattens a space-indented file. Raised with @EduardMe; the changelog tells users to indent with tabs.

## Second fix: a missing argument, hidden by a computed function name

```js
const insertFunc = lastPara.type === 'separator' ? `insertTodoBeforeParagraph` : `insertParagraphAfterParagraph`
note[insertFunc](content, lastPara)   // <- one of these takes THREE arguments
```

`insertParagraphAfterParagraph(content, otherParagraph, paragraphType)` was being called with two arguments, so `paragraphType` arrived as `undefined`. The computed-name form is what hid it: the two possible targets have different arities, so nothing at the call site — and nothing Flow could see — flagged the missing argument. Now split into explicit branches, passing `'text'` to match every other insertion path in the same function.

Full disclosure on how this landed: I reverted this once, wrongly. I attributed observed indent-flattening to it, but the log showed the branch had never executed (the run took the `shouldInsertAtTop` path through `addParagraphBelowHeadingTitle`), and the flattening was in fact NotePlan discarding space-indentation. With that explanation gone there is no evidence against the argument, and passing `undefined` for a required API parameter is a real defect, so it is re-applied here.

**Needs a smoke test with the *tasks to top* setting OFF**, since that is the only way to reach this branch.

## Release

Version bumped **1.2.9 → 1.3.0** with a changelog entry, ready to release separately.


dwertheimer:
## Validated end-to-end in NotePlan (2026-08-02)

Tab-indented note, sorted via `Sort tasks under heading`:

```
# tab sort test          # tab sort test
## Test Sort             ## Test Sort
* zebra task       ->    * apple task !!
* apple task !!          →	* child of apple
	* child of apple     →		* grandchild of apple
		* grandchild     * mango task
* mango task             * zebra task
```

Confirmed by reading the file bytes on disk afterwards: `<TAB>` and `<TAB><TAB>` are preserved, **not** converted to spaces. That matters because NotePlan discards space-indentation, so a spaces round-trip would silently flatten the note on the next sort. This one is idempotent.

Log evidence for each fix:

- **Nesting:** `Open Tasks:3` (was 5 pre-fix), with `apple → child → grandchild` nested two deep.
- **Deletion set:** `Queuing 3 open tasks` → `Deleting 5 tasks from note`. The extra 2 are apple's descendants. A one-level `children.forEach()` would have deleted 4 and left a duplicate in the note — this is the round-trip invariant the unit tests pin, firing for real.
- **Setting fix:** `sortTasksUnderHeading: interleaveTaskTypes=false (from setting)`, then `shouldInsertAtTop=false`.
- **Missing argument:** `insertTodos note.insertParagraphAfterParagraph "Test Sort"` — reached for the first time, anchored on the heading paragraph.

That last one also disproves a theory I had floated in review: `getBlockUnderHeading` returns 1 fresh paragraph after the deletions, so `note.paragraphs` is not stale and the anchor is correct.

Minor cosmetic note, unrelated to these fixes: blank lines within the sorted block are not preserved across the delete/re-insert. Pre-existing behaviour as far as I can tell; flagging it rather than changing it.

PR #772, opened 2026-08-04
CI: enforce zero Flow errors (replace the baseline ratchet with a strict check)

dwertheimer:
## What

Type errors now block your work at two points:

- **Locally** — a pre-push hook refuses the push and prints the errors, so you find out in seconds.
- **In CI** — a Flow job runs on **every branch push**, not just PRs, and fails on any error.

The hook installs itself via a `prepare` script on `npm install`. No new dependencies. To bypass in an emergency: `git push --no-verify` (CI will still fail on it).

## Why

The repo is now at zero Flow errors, and this is the moment to lock that in. Without a gate it drifts straight back — type debt only ever accumulates quietly.

Flow types used to be a real tax: annotating a large codebase by hand is slow, which is why the errors were allowed to pile up in the first place. **That has changed.** With AI assistance, writing and maintaining correct types is close to free — inferring them, propagating them, and fixing the fallout when a signature changes is now cheap. Staying at zero costs far less than getting to zero did.

It also matters more now that AI writes a good share of this code. Type checking is one of the few genuinely mechanical checks on correctness — one that neither a human reviewer nor the model reliably applies to itself. Cleaning up the existing errors turned up real bugs hidden behind suppressions: calls to methods that did not exist and threw every time, string operations on non-strings, null dereferences. Flow would have caught all of them at authoring time.

Enforcing it on every push keeps that guarantee honest for humans and AI alike.

## Trade-off

A branch that introduces a single Flow error is blocked until it is fixed. That is the intent, but it changes day-to-day workflow for everyone — hence a standalone PR.

Flow runs in its own workflow so it can trigger on every branch; jest and the plugin builds keep their existing narrower trigger, since those are much more expensive to run.

`node scripts/flow-report.js` remains available locally for a per-plugin breakdown, and `--plugin <name>` lists individual error sites.

@jgclark @eduardme — flagging you both, since this changes the rules for all contributors.


jgclark (approved):
I can't comment on the specifics of the CI, but I approve the principle, so I'm happy to try going for it.

dwertheimer:
Thanks @jgclark OK I am going to merge it. @EduardMe Just FYI. Turning on a hard-reject for Flow errors.

dwertheimer:
@jgclark @EduardMe Make sure you pull the latest code after this PR merge

